PR #5, opened 2022-02-18
Add another state listener to the Slack notification

chiaberry:
We have some flows failing, but the failures aren't being reported to Slack. I want to try adding another state to listen for to see if it will show up in slack. 

https://docs.prefect.io/api/latest/engine/state.html#triggerfailed

### DIFF
--- a/flows/MDS/bird.py
+++ b/flows/MDS/bird.py
@@ -17,7 +17,7 @@ from datetime import datetime, timedelta
 from prefect import Flow, task
 from prefect.storage import GitHub
 from prefect.run_configs import UniversalRun
-from prefect.engine.state import Failed
+from prefect.engine.state import Failed, TriggerFailed
 from prefect.schedules import Schedule
 from prefect.schedules.clocks import CronClock
 from prefect.backend import get_key_value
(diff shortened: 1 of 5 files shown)


flows/MDS/bird.py
```diff
@@ -29,9 +29,8 @@
 current_environment = os.getenv("PREFECT_CURRENT_ENVIRONMENT", "staging")
 
 # Set up slack fail handler
-handler = slack_notifier(only_states=[Failed])
+handler = slack_notifier(only_states=[Failed, TriggerFailed])
```
chiaberry, line 32: I would like to try to see if this state happens when the flows don't start running at all.

Charlie-Henry: [It looks like](https://docs.prefect.io/api/latest/engine/state.html#state-2) `TriggerFailed` is a substate of the `Failed` state. Possible there's a bug and this isn't working how they thought?
```
my_state = TriggerFailed()
my_state.is_failed() # returns True
```

The silent error in the task from last night is actually technically in a `pending` state?
Screen Shot 2022-02-18 at 3 43 22 PM

mateoclarke (approved):
I'm good with this but seeing Charlie's comment makes me wonder if it would have any effect.

Charlie-Henry:
I guess silent failures haven't been an issue for us now that we aren't getting these failures anymore but I saw this [issue](https://github.com/PrefectHQ/prefect/issues/5652) mentioned in the Prefect slack about this exact thing. 

[Slack thread](https://prefect-community.slack.com/archives/CL09KU1K7/p1649265574670459)

PR #6, opened 2022-02-28
Add flow for knack banner HR app

chiaberry:
The long awaited Flow for the [atd-knack-banner](https://github.com/cityofaustin/atd-knack-banner) scripts.

In short, the script compares records from the Banner endpoint to the CTM email excel spreadsheet and updates the knack HR app according to differences between the three. 

This flow will then take the outputted logs from the script and then email that to an email address that is defined in the prefect cloud key value store. 

The flows will register with the staging agent and the production agent, as is our current configuration.


docs/01-basics.md
```diff
@@ -136,99 +135,29 @@ you have the option to register the flow directly
...
-     $ prefect register --project my-project -p myflows/ --watch
-
-   Register a flow found in `flow.py` and disable its schedule.
-
-     $ prefect register --project my-project -p flow.py --no-schedule
+$ prefect register --project MDS --path <your file path>
+```
 
+```bash
+$ prefect register -h
```
frankhereford, line 158: Thanks for this change - this won't get stale as they add arguments to the prefect command.

docs/01-basics.md
```diff
@@ -292,7 +221,7 @@ before execution. While downloading the file is
 not necessary for us, it makes it very helpful when
 running in remote/cloud environments.
 
-For us to use the class, we simply declare the
```
frankhereford, line 295: 👍

flows/knack/knack_banner.py
```diff
@@ -0,0 +1,127 @@
...
+from prefect.engine.state import Failed, TriggerFailed, Retrying
+from prefect.schedules import Schedule
+from prefect.schedules.clocks import CronClock
+from prefect.backend import get_key_value
+from prefect.triggers import all_successful
+
+from prefect.utilities.notifications import slack_notifier
+from prefect.tasks.notifications.email_task import EmailTask
+
+current_environment = os.getenv("PREFECT_CURRENT_ENVIRONMENT", "production")
```
frankhereford, line 30: Would this be an opportunity to define the "default" environment to be `staging` to have a fail-to-safe type of setup?

flows/knack/knack_banner.py
```diff
@@ -0,0 +1,127 @@
...
+    state_handlers=[handler],
+    slug="knack-banner",
+)
+def knack_banner_update_employees():
+    response = (
+        docker.from_env()
+        .containers.run(
+            image=docker_image,
+            working_dir="/app",
+            command=f"./atd-knack-banner/update_employees.py",
```
frankhereford, line 56: This couldn't be less of a thing, and no harm as is, but would you want to not make this an f-string to match other non-interpolated strings?

flows/knack/knack_banner.py
```diff
@@ -0,0 +1,127 @@
...
+    logger.info(response)
+    return response
+
+
+# Configure email task
+email_task = EmailTask(
+    name="email_task",
+    subject="Knack HR updates from Banner",
+    email_to=environment_variables["EMAIL_ADDRESS"],
+    email_to_cc="chia.berry@austintexas.gov",
```
frankhereford, line 75: Do you want your email in this as it gets into `main`?

flows/knack/knack_banner.py
```diff
@@ -0,0 +1,127 @@
...
+            environment=environment_variables,
+            volumes=None,
+            remove=True,
+            detach=False,
+            stdout=True,
+        )
+        .decode("utf-8")
+    )
+    logger = prefect.context.get("logger")
+    logger.info(response)
```
frankhereford, line 66: This is a naïve question, but does this logger.info() method provide a debugging aid or does the message that gets emitted get handled to become visible via prefect's UI or slack, maybe?

flows/test/dependent_flows.py
```diff
@@ -0,0 +1,102 @@
+#!/usr/bin/env python
+
+"""
+Name: Dependent flows testing
+Description: The purpose of this file is to serve as a template to
```
frankhereford, line 5: 🙏 Thank you for making a template, this is super helpful to say the least!

flows/test/dependent_flows.py
```diff
@@ -0,0 +1,102 @@
...
+from prefect.tasks.prefect import create_flow_run, get_task_run_result
+from prefect.storage import GitHub
+from prefect.run_configs import UniversalRun
+from prefect.engine.results import PrefectResult
+from prefect.backend import get_key_value
+
+from prefect.tasks.notifications.email_task import EmailTask
+
+# First, we must always define the current environment, and default to staging:
+current_environment = os.getenv("PREFECT_CURRENT_ENVIRONMENT", "staging")
```
frankhereford, line 25: 👍

flows/test/dependent_flows.py
```diff
@@ -0,0 +1,102 @@
...
+    smtp_server=email_config["smtp_server"],
+    smtp_port=email_config["smtp_port"],
+    smtp_type=email_config["smtp_type"],
+    attachments=None
+)
+
+
+@task(log_stdout=True)
+def format_dict(flow_data):
+    print(f"Got: {flow_data!r}")
```
frankhereford, line 55: I did a quick search for the `!r` syntax, and I'm wondering, is this a typo, or a future TIL for me.

frankhereford (approved):
Thank you for this new flow both for what it does but also as a demonstration of how to get stuff working. As we discussed, there isn't an obvious way to execute as we don't have development environments for all the systems this integrates with. As such, I have only gone over the code carefully with an eye for understanding the process intends to do and the dependencies of the steps along the way. I think this all looks good - I have a few questions here and there, but nothing that would hold up merging.

flows/knack/knack_banner.py
chiaberry: the default environment usually is staging, i changed it here to be production because if it tries to run on staging its fails right out the gate.

flows/knack/knack_banner.py
chiaberry: it shows up in prefect's UI if use this logger

flows/test/dependent_flows.py
chiaberry: this is a giant typo.

flows/knack/knack_banner.py
```diff
@@ -0,0 +1,127 @@
...
+    get_data_flow_run_id = create_flow_run(flow_name=get_data_flow.name)
+    script_result = get_task_run_result(
+        get_data_flow_run_id, task_slug="knack-banner-copy"
+    )
+    formatted_data = format_email_body(script_result)
+    send_email_flow.chain(script_result, formatted_data, email_task(msg=formatted_data))
+
+
+if __name__ == "__main__":
+    send_email_flow.run()
```
Charlie-Henry, line 127: How does the `send_email_flow` flow get triggered? I see the scheduled flow and task running each day for `atd_knack_banner` but how does that talk to the second `send_email` flow? 

Very cool thanks for sharing with me!

Charlie-Henry (approved):

flows/knack/knack_banner.py
chiaberry: good question. i keep going back and forth, ill put it in an environment variable.

flows/knack/knack_banner.py
chiaberry: There's some magic Sergio wrote here in the github actions: https://github.com/cityofaustin/atd-prefect/blob/main/.github/workflows/prefect_helper.sh#L79

mateoclarke (approved):
🚀 🤘 ✨

flows/knack/knack_banner.py
```diff
@@ -0,0 +1,127 @@
...
+    schedule=Schedule(clocks=[CronClock("45 13 * * *")])
+) as get_data_flow:
+    email_data = knack_banner_update_employees()
+
+
+with Flow(
+    f"send_hr_email_{current_environment}",
+    storage=GitHub(
+        repo="cityofaustin/atd-prefect",
+        path="flows/test/knack_banner.py",
```
mateoclarke, line 113: Is this path `flows/test/knack_banner.py` right? I don't see a knack_banner.py script in the test directory. Should it be in the knack directory?

mateoclarke (approved):

PR #7, opened 2022-03-01
change parking to the end of the month

Charlie-Henry:
I ran at 6:36PM on the last day of February Austin time and it was searching the March folder in S3 which would only happen if `datetime.now()` was returning time in UTC. This defaults to whatever your local machine is set to. 

This change will now will tell Prefect to search for and download data for the current and previous month if the day is between the 27th and 7th days (near the start/end of each month). Purpose is to cut down on unnecessary processing of data but that may be needed on the last couple days of the month.


mateoclarke (approved):
I don't see the local timezone change you mentioned in the update but the range change looks good and make sense for practical purposes. I would suggest adding a one line comment to the code just in case anyone stumbles here and wonders about the seemingly arbitrary range.

PR #10, opened 2022-05-06
Use correct github ref for knack banner flow

chiaberry:
The flow was erroring immediately because it was looking for the file in github in the production branch, based on the current_environment. I am changing the ref to be main, regardless of environment file. 

I'm also correcting a mistake that Mateo pointed out when he reviewed this file earlier, that I neglected to fix.

### DIFF
--- a/flows/knack/knack_banner.py
+++ b/flows/knack/knack_banner.py
@@ -97,7 +97,7 @@ with Flow(
     storage=GitHub(
         repo="cityofaustin/atd-prefect",
         path="flows/knack/knack_banner.py",
-        ref=current_environment.replace("staging", "main")
+        ref="main"
     ),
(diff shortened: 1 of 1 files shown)


Charlie-Henry:
I think this is supposed to point to `staging` instead of `production`. That should fix your issue and then if we pull this into production it'll still work. https://github.com/cityofaustin/atd-prefect/blob/3ec312f7ebcdd1b408465cde6a66741f42c0708b/flows/knack/knack_banner.py#L30

chiaberry:
If i default it to staging, then this line messes up
`environment_variables = get_key_value(key=f"atd_knack_banner_{current_environment}")` because the staging knack HR app is behind the production one and has been throwing errors. 

I thought we were going to deprecate the production branch, but theres a good chance I'm remembering it wrong.

frankhereford (approved):

PR #11, opened 2022-05-06
Moped Finance Data Socrata Flow

Charlie-Henry:
Script moves data from an S3 bucket to a Socrata dataset. This is used in Moped. Uses [this script](https://github.com/cityofaustin/atd-finance-data/blob/production/s3_to_socrata.py).

Issues: 
- https://github.com/cityofaustin/atd-data-tech/issues/7931
- https://github.com/cityofaustin/atd-data-tech/issues/8809

This has been running in `test` now for a while so this is for moving this into production.

Moved latest changes into `staging`: https://cloud.prefect.io/flow/4526caed-62e3-4f3e-aa00-40da176974b5

### DIFF
--- a/flows/parking/parking_data_reconciliation.py
+++ b/flows/parking/parking_data_reconciliation.py
@@ -1,0 +1,219 @@
+#!/usr/bin/env python
+
+"""
+Name: Parking Data Reconciliation Flows
+Description: Parse Fiserv emails then upsert the CSVs to a postgres DB.
+    Grab the payment data retrived from flowbird and upsert tha to a postgres DB.
+    Then, compare them before uploading the data to Socrata.
+Schedule: "30 5 * * *"
+Labels: atd-data02, parking
+"""
+
+import os
+
+import docker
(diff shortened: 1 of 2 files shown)


flows/parking/parking_data_reconciliation.py
```diff
@@ -0,0 +1,277 @@
+#!/usr/bin/env python
+
+"""
+Name: Parking Data Reconciliation Flows
+Description: Parse Fiserv emails then upsert the CSVs to a postgres DB.
+    Grab the payment data retrived from flowbird and upsert tha to a postgres DB.
+    Then, compare them before uploading the data to Socrata.
+Schedule: "30 5 * * *"
```
frankhereford, line 8: Super nothing, but I think this schedule isn't the same as the cron schedule that the flow registers with Prefect.

Charlie-Henry: thanks Frank, I always forget to update the template...

PR #12, opened 2022-05-12
Finance Production was failing

Charlie-Henry:
This flow was still looking for `atd_finance_staging` because I was tagging this with `test`. Changed to be `current_environment`. 

If I understand this correctly, that the `atd-data02` production agent will have an environment variable called  PREFECT_CURRENT_ENVIRONMENT and this will work? That seems to be the pattern that we've implemented so far.

```
current_environment = os.getenv("PREFECT_CURRENT_ENVIRONMENT", "staging")
```

### DIFF
--- a/flows/moped/finance_data.py
+++ b/flows/moped/finance_data.py
@@ -95,7 +95,7 @@ with Flow(
     ),
     # Run config will always need the current_environment
     # plus whatever labels you need to attach to this flow
-    run_config=UniversalRun(labels=["test", "atd-data02"]),
+    run_config=UniversalRun(labels=[current_environment, "atd-data02"]),
     schedule=Schedule(clocks=[CronClock("30 12 * * *")]),
 ) as flow:
     flow.chain(pull_docker_image, s3_to_socrata)


mateoclarke (approved):
👍

mateoclarke:
I'm going to tag this item for discussion in our next Prefect Huddle.

frankhereford (approved):

Charlie-Henry:
Hmmm this is still failing, curious what you all discussed on Tuesday: 
@frankhereford @mateoclarke 

```
[19 May 2022 10:53am]: Failed to load and execute Flow's environment: 
ValueError("Flow 'atd_finance_production' not found in file. Found flows:\n- 'atd_finance_staging'")
```

Looks like it was using the right commit, too:
```
Flow successfully downloaded. Using commit: b518556dc0510e060d218754c237fae1f5bb9c16
```

chiaberry:
For what its worth, the `test` `atd-data02` agent didn't have a `PREFECT_CURRENT_ENIVRONMENT` set, so I set it to production. I can remove it, or reset it to "staging" if that is what flows on that agent expect.

Charlie-Henry:
The `test` agent should either have no `PREFECT_CURRENT_ENIVRONMENT` variable set or set to `staging`? The parking ETL running on test is expecting it to default to `staging`. Not sure if that's the pattern we want or not?

PR #17, opened 2022-07-18
Logging Level Issue

Charlie-Henry:
I have an issue with the Parking flows that sometimes causes the passport_DB task to fail. I believe it is something to do with one of the .json files I have downloaded from Passport but they get overwritten during each execution so the issue eventually fixes itself. This makes it harder to debug what the issue is without some logs.

Here's one run that [failed](https://cloud.prefect.io/dts/flow-run/50e1015e-4e36-42a4-a940-d090bb4b3353?logs)

When a flow fails, it just seems to log the `ERROR` level of logging and none of the `INFO` or `DEBUG`, which is where I have been logging which file was loaded and could point us to what is failing.

I'm trying adding `log_stdout=True` to see if this could be the issue?
https://docs.prefect.io/core/concepts/logging.html#logging-stdout

### DIFF
--- a/flows/parking/atd_parking_data_meters_txn_history.py
+++ b/flows/parking/atd_parking_data_meters_txn_history.py
@@ -79,6 +79,7 @@ start_date = get_start_date(prev_execution_date_success)
     timeout=timedelta(minutes=60),
     retry_delay=timedelta(minutes=5),
     state_handlers=[handler],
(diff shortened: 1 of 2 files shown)


johnclary (approved):
🛶

mateoclarke (approved):
🪵

PR #20, opened 2022-07-21
Moped test API and DB tasks

mddilley:
This is the sibling to https://github.com/cityofaustin/atd-moped/pull/725. I set the version of zappa in the dependencies to [the version pinned in the current deployment scripts](https://github.com/cityofaustin/atd-moped/blob/main/moped-api/requirements/production.txt).


frankhereford (approved):
Looks awesome - thank you @mddilley!

PR #25, opened 2022-07-27
Dockerized Agent system for Microstrategy

frankhereford:
@Charlie-Henry This is all yours if it's helpful. Thank you again for what you taught me today; that was a really valuable meeting to me! 🐳🚀


frankhereford:
👋 Charlie - Do you think we should close this PR? No worries, I just came across it yesterday, and I think we've gone another way. What do you think is best?

frankhereford:
Awesome - thanks Charlie.

PR #26, opened 2022-07-27
Orchestrate Prefect Docker agent for atd-data03

frankhereford:
## Prefect Agent which is able to pull docker images and run flows in them

During a conversation with @Charlie-Henry, he demonstrated the use of a Docker Registry as a viable mechanism to convey flows and their related environments from a developer's machine to a production deployment. Using such a system, a developer can build the environment they wish a flow to run into a docker container, upload it to a registry, and register a flow that specifies the image and version to pull from the registry in which to execute their task. 

After this demonstration, I offered to stand up an agent on `atd-data03` which would receive such flows and be able to handle them.

This PR includes a Dockerfile to define a docker container within to run such an agent and includes the bind mounting tricks to allow the agent in that docker container to pull and execute its own docker images and containers, but have that actually happen at the host VM level. It will also provide for the agent to be restarted on system reboot through `docker-compose`'s features.

In short, I believe this will stand up a viable target for our team's Prefect flow developers to push dockerized flows and have them run. It will run with the labels: `docker`, `production` and `atd-data03`.


agents/atd-data03-docker/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM ubuntu:22.04
+ENV TZ="America/Chicago"
+RUN apt-get update
+RUN apt-get update && \
+    apt-get install -yq tzdata && \
+    ln -fs /usr/share/zoneinfo/America/Chicago /etc/localtime && \
+    dpkg-reconfigure -f noninteractive tzdata
+RUN apt-get upgrade -y
+RUN apt-get install -y vim magic-wormhole aptitude python3-pip docker.io black
+COPY ./requirements.txt /root/requirements.txt
```
Charlie-Henry, line 10: Thanks for making this Frank! 
Curious, I see we're our own `requirements.txt` file for Prefect, are we using that over the [base prefect images](https://hub.docker.com/r/prefecthq/prefect) because we're worried that a prefect update could cause things to break?

Charlie-Henry (approved):

agents/atd-data03-docker/Dockerfile
frankhereford: This is a good question - this requirements.txt actually is used to put python libraries (such as `prefect`) on the host that runs the agent, which is going to be a docker container. That agent itself will fire up the docker image you create and push to the registry during flow registration. It's two layers deep of docker containers logically, but in practice, they all are running as siblings on the host container thanks to the passing down of the docker daemon control socket.

mddilley (approved):
Looks good to me - do we have an example of a flow that is uploaded to a registry and handled this way? Sorry it took so long to look at this, and apologies if I am behind the curve here. 🙏

frankhereford:
@chiaberry I think this is ready to merge, but I wanted you to know since it's so old. That cool?

chiaberry:
if its in use, for sure merge it

PR #27, opened 2022-07-27
Microstrategy Reports Docker Agent Flow

Charlie-Henry:
This Prefect flow is a simple ETL for downloading reports from microstrategy to a S3 bucket. It was created for https://github.com/cityofaustin/atd-data-tech/issues/9645

### Tasks:
1. Get microstrategy connection (need to be on city network via VPN or use something on-prem)
2. Connect using boto3 to our S3
3. Download report to a dataframe (for each report_id)
4. Send report dataframe to S3 bucket (for each report_id)

### Parameters
They can be configured in Prefect Cloud, only one is defined as default in this PR `"2020 Bond Expenses Obligated"`:
- report_ids
- report_names (must be unique for each report)

### Docker
This flow uses `DockerRun` combined with `Docker` storage. It must be run on a [Docker Agent](https://docs-v1.prefect.io/orchestration/agents/docker.html#agent-configuration), which @frankhereford has set up for me on `atd-data03`.

When this flow is registered, the  `python_dependencies=["mstrio-py", "pandas", "boto3"]` are installed with the [base Prefect image](https://hub.docker.com/r/prefecthq/prefect) for the given Python version we're running in a docker container and then this is pushed up to our atddocker account. The Docker agent will then pull down this docker container and use it to run our tasks 🪄✨.

Using `python_dependencies` is pretty painless but you can also provide your own Dockerfile using `dockerfile = "path/to/Dockerfile"`.

***

It is currently running [here](https://cloud.prefect.io/dts/flow/19bd62c9-a7d5-44c9-a796-d1ea00bd7f41?overview) using labels:`atd-data03` `production` `docker`


frankhereford (approved):
🚀

Thanks @Charlie-Henry! Let me know if there are any snags with the production environment, and I will seek approval and merge [PR 26](https://github.com/cityofaustin/atd-prefect/pull/26) as soon as I'm able.

flows/microstrategy/mstro_to_s3.py
```diff
@@ -0,0 +1,144 @@
...
+    flow.add_task(names)
+
+    # 1. Get microstrategy connection
+    conn = connect_to_mstro()
+
+    # 2. Connect using boto3 to our S3
+    s3 = connect_to_AWS()
+
+    # 3. Download report to df (for each report_id)
+    df = download_report.map(ids, unmapped(conn))
```
johnclary, line 129: if we add more reports to this script, do we really want to process multiple reports in go? might be better to take the report name, or id or whatever as an argument so that they can be scheduled separately and not fail together?

flows/microstrategy/mstro_to_s3.py
```diff
@@ -0,0 +1,144 @@
...
+# To find report ID, go to the report in Microstrategy then:
+## Go to Tools > Report Details Page or Document Details Page.
+## Click Show Advanced Details button at the bottom
+# Report name should be unique as it is the file name in the S3 bucket
+REPORTS = [["2020 Bond Expenses Obligated", "85A9E0A049F06D98AF1CF3BE8CDA9394"]]
+
+## AWS Credentials
+AWS_ACCESS_ID = environment_variables["AWS_ACCESS_ID"]
+AWS_PASS = environment_variables["AWS_PASS"]
+BUCKET_NAME = environment_variables["BUCKET"]
```
johnclary, line 45: might be worth adding a copy of the env vars object you've got stored in prefect to 1Pass

flows/microstrategy/mstro_to_s3.py
```diff
@@ -0,0 +1,144 @@
...
+
+    # 3. Download report to df (for each report_id)
+    df = download_report.map(ids, unmapped(conn))
+
+    # 4. Send file to S3 bucket (for each report_id)
+    report_to_s3.map(df, names, unmapped(s3))
+
+flow.storage = Docker(
+    registry_url="",  # Needs to be empty string for pushing to dockerhub
+    image_name="atddocker/atd-microstrategy",
```
johnclary, line 136: so, there is no Dockerfile? this flow, when registered, did the work of `docker build` and `docker push` and automatically built on top of prefect's base image?

johnclary (dismissed):
@Charlie-Henry this is super cool to see! Apart from processing mstr data (awesome), there are cool prefect patterns here for working with S3. goood stuff.

This flow definitely needs a readme, which would provide the business context for what these reports are, who relies on them, and describes the other pieces of this pipeline, e.g., how does the S3 data get to PowerBI?

mateoclarke (approved):
🚀

frankhereford (approved):
time

Sorry for the double approval! Github brought me here and showed me your very recent update re the timeout. Thanks!

flows/microstrategy/mstro_to_s3.py
Charlie-Henry: Prefect does this a little strange, `ids` looks like a list of `ids` but the flow is behaving more like a loop where each `id` is sent to the task individually. It's called "[mapping](https://docs-v1.prefect.io/core/concepts/mapping.html)". If a `download_report` task fails, it should continue onto to the next report and not run any subsequent tasks for that report.

flows/microstrategy/mstro_to_s3.py
Charlie-Henry: You can supply a Dockerfile if that is desired. It start with the Prefect base images on dockerhub for the python version you are running, then install the `python_dependencies`.

When you register this flow it will `docker build`, and it should `docker push` to dockerhub, although I've only seen this happen once when I originally registered the flow. I've had to manually `docker push` after that.

Each time the flow runs on a docker agent, it does a `docker pull` so if your image changes, it should see those changes and apply them.

PR #28, opened 2022-08-01
Rework VZ import script

frankhereford:
# Primary Goal

**Make VZ ETL script overwrite safe and allow for backfilling without losing user-entered data**

## Goals

- [x] Utilize Postgres typing rigidity and data comparison functionality to increase the reliability of the import process.
- [x] Implement the "Ideal ETL" as defined in 7/12 whiteboarding session, see [#8746](https://github.com/cityofaustin/atd-data-tech/issues/8746#issuecomment-1182005821)'s comment.
- [x] Improve documentation around the ETL process
- [x] More tightly integrate the Prefect system for logging and reporting of import volumes
- [x] Eliminate differences in the way that crash vs non-crash records are handled

## Associations

This PR hopes to close https://github.com/cityofaustin/atd-data-tech/issues/8746.
This PR has a sibling in the VZ repo: [PR 1096](https://github.com/cityofaustin/atd-vz-data/pull/1096).

## TODO

- [x] Utilize conflict resolution system
- [x] `Insert`s
- [x] `Update`s
- [x] Skip updates if no change is needed
- [x] Move SQL generation routines into library for readability
- [x] Unify into KV store
- [ ] ~📒 Document in Gitbook.~
  - [x] The more I leave this here, the more I realize I don't have a ton to write yet. I am curious if you agree that the comments in the flow and the main tasks are sufficient. 
- [x] 🗯 Comments in code
- [x] 🥼Figure out a testing scheme .. 
  - Proposing to have a meeting with demonstrations and prepared fixtures as well as fresh extracts from CRIS yet untested in lieu of individual, self-driven testing, although the later is of course welcome.
- [ ] ~Consider escaped SQL values pursent to SQL injection rejection~
- [ ] ~Add examples of formed SQL fragments and queries in `sql.py`~
  - [x] It's easy enough to add a one-off print() statement, and they are unweildly in a comment (100s of lines long)

## Future Considerations
- [ ] `charges` table

## Testing

* I would be excited to facilitate anyone who wants to spin up a local container of the VZ prefect agent docker image. 
* Realistically though, this is a real bear to test. Roughly the steps would be:
  * Build local VZ environment, run it (db, graphql-engine, vze)
  * Export out a known backup from a point in time, load it
  * Run the script, or register and invoke via cloud for bonus points
  * Inspect the data in the CSV, the `import` schema, and the involved export tables for correctness
* As a result, I'll propose that I demonstrate it running in my local environment with a medium size interval of export backfill data in a meets call or similar. I could record it and take advantage of post processing to ⏩ for the boring parts.

### Thoughts

* Just like the previous script which made tons of graphql requests to the AWS hosted VZ endpoint, this will make many queries and transactions to the VZ db itself. Minimizing network latency between the script & the postgres server will directly & linearly correlate to overall processing speed.
* Because this throws an exception on query errors, it will more aggressively warn if there are key errors or violations of unique constraints. This is a good thing.
* This program is unable to transfer in data if a field is not existing in the target tables, therefore, the previous ETL's consideration of PII columns is dropped.


frankhereford:
Little test case to test equality testing with regard to null values:

```
create table alpha (name character varying, food character varying);
create table beta (name character varying, food character varying);

insert into alpha (name, food) values ('frank', 'tacos');
insert into beta (name, food) values ('frank', 'tacos');
insert into alpha (name, food) values ('stacy', 'burgers');
insert into beta (name, food) values ('stacy', 'pasta');
insert into beta (name, food) values ('steve', 'pot luck');
insert into alpha (name, food) values ('mateo', null);
insert into beta (name, food) values ('mateo', null);

delete from alpha where name = 'stacy' and food = 'pasta';

select alpha.name as alpha_name, alpha.food as alpha_food, beta.food as beta_food,
  (
    alpha.food = beta.food
    or
    (alpha.food is null and beta.food is null)
  ) as is_equal
from alpha
left join beta on (alpha.name = beta.name);
```

flows/vision-zero/cris_import/cris_import.py
```diff
@@ -294,81 +320,246 @@ def remove_archives_from_sftp_endpoint(zip_location):
...
+
+                # For future reporting and debugging purposes: Use SQL to query a list of 
+                # column names which have differing values between the import and target records.
+                # Return these column names as an array and display them in the output.
+                changed_columns = util.get_changed_columns(pg, column_aggregators, output_map, table, linkage_clauses, public_key_sql, DB_IMPORT_SCHEMA)
+                if len(changed_columns):
+                    print("Changed Columns: " + str(changed_columns["changed_columns"]))
+
+                # Using all the information we've gathered, form a single SQL update statement to update the target record.
+                update_statement = util.form_update_statement(output_map, table, column_assignments, DB_IMPORT_SCHEMA, public_key_sql, linkage_sql)
```
frankhereford, line 504: TODO: This function should be extended to take `changed_columns` and query out the starting values of the columns in question in the input and target record. Printing these would be invaluable in tracking updates in the prefect logs.

flows/vision-zero/cris_import/lib/sql.py
```diff
@@ -0,0 +1,460 @@
...
+    """
+
+    cursor = pg.cursor(cursor_factory=psycopg2.extras.RealDictCursor)
+    cursor.execute(sql)
+    input_table_column_types = cursor.fetchall()
+
+    # TODO figure out how to .map() this
+    input_column_names = []
+    for result in input_table_column_types:
+        input_column_names.append(result["column_name"])
```
mddilley, line 273: [List comprehensions](https://docs.python.org/3/tutorial/datastructures.html#list-comprehensions) are fun for this but this might be more readable as is 👍

flows/vision-zero/cris_import/lib/sql.py
```diff
@@ -0,0 +1,460 @@
...
+    sql = "select (" + " and ".join(column_comparisons) + ") as skip_update\n"
+    sql += f"from public.{output_map[table]}\n"
+    sql += (
+        f"left join {DB_IMPORT_SCHEMA}.{table} on ("
+        + " and ".join(linkage_clauses)
+        + ")\n"
+    )
+    sql += f"where {public_key_sql}\n"
+
+    cursor = pg.cursor(cursor_factory=psycopg2.extras.RealDictCursor)
```
mddilley, line 93: TIL about psychopg2's [RealDictCursor](https://www.psycopg.org/docs/extras.html#real-dictionary-cursor) 👍

mddilley (approved):
Thanks for the demo this morning, and I really like the way this code is organized to keep the path through the script straightforward to see. The helpers tuck away the harder to understand SQL parts and isolate them into some really nice packages. 🎁 Awesome to see this implemented with pgfutter as well!

patrickm02L (approved):
Approving based on 9/14/22 demo. Good stuff. 🤘

flows/vision-zero/cris_import/lib/mappings.py
```diff
@@ -0,0 +1,94 @@
...
+            "prsn_occpnt_pos_id",  # key column (but why?)
+            "injury_severity",
+            "prsn_age",
+        },
+        "atd_txdot_primaryperson": {
+            "crash_id",  # key column
+            "unit_nbr",  # key column
+            "prsn_nbr",  # key column
+            "prsn_type_id",  # key column (but why?)
+            "prsn_occpnt_pos_id",  # key column (but why?)
```
frankhereford, line 84: These 'but why?' comments are for columns which are used in the primary key index in the VZ database, but do not appear to be required to uniquely define a `person` or `primaryperson`. I think we get by with this misconfigured PK in our DB because we only create records when CRIS does upstream, and they have correct indexing, which we inherit the effect of.

chiaberry (approved):
Thanks for the demo, I like the splitting out of sql into their own functions, so its easier to follow.

frankhereford:
# Deployment

I'm going to roll this work out in a few discrete steps over the next few days. cc/ @patrickm02L @chiaberry 

The steps are:

# Deployment Steps

## Day 1

* [x] Turn off airflow DAGS `atd-visionzero_download_production` and `atd_visionzero_hasura_input_production.` No staging or development DAGS for these processes were turned up.
* [x] Observe an extract sitting in the SFTP endpoint, having an md5 sum starting with `d0fb29`
* [x] Disable `/root/automated_tasks/extract_s3_daily.sh` in `root`'s crontab.
* [x] Build the new VZ dockerized agent image (`docker-compose build --no-cache`) on data03
* [ ] Restart it and observe that flows are being processed.

## Day 2

* [ ] Backup or confirm backup the VZDB as close to the time of execution
* [ ] Run the script, watching the entire output and trapping to a log with `tee`
* [ ] Check the status of the VZDB, VZE and mitigate any problems that come up
* [ ] Be sure to check status of SFTP endpoint

## Day 3
* [ ] Backup or confirm backup the VZDB
* [ ] Register the script and run manually via the cloud UI
* [ ] Prepare schedule, look at logs in cloud UI
* [ ] Be sure to check status of SFTP endpoint

## Day 4
* [ ] Run manually again
* [ ] Enable schedule

## Day 5
* [ ] Observe automated execution

PR #36, opened 2022-08-17
Don't bind mount the `/tmp` directory into the container

frankhereford:
Sibling PR: https://github.com/cityofaustin/atd-moped/pull/762


mddilley (approved):
🚀

PR #2, opened 2022-02-09
[WIP] Add AFD ETL for Vision Zero Database

mateoclarke:
Closes https://github.com/cityofaustin/atd-data-tech/issues/8159

Its possible that some of this should live in the VZ Repo but we can refactor that in the future.
https://github.com/cityofaustin/atd-vz-data/tree/master/atd-etl

PR #38, opened 2022-08-30
Add EMS/AFD flows for VZD

mateoclarke:
Replaces https://github.com/cityofaustin/atd-prefect/pull/2/files

Companion to https://github.com/cityofaustin/atd-vz-data/pull/1095


flows/vision-zero/Dockerfile
```diff
@@ -6,7 +6,7 @@ RUN apt-get update && \
     ln -fs /usr/share/zoneinfo/America/Chicago /etc/localtime && \
     dpkg-reconfigure -f noninteractive tzdata
 RUN apt-get upgrade -y
-RUN apt-get install -y vim magic-wormhole aptitude python3-pip docker.io rsync p7zip-full black poppler-utils tesseract-ocr
+RUN apt-get install -y vim magic-wormhole aptitude python3-pip docker.io rsync p7zip-full black poppler-utils tesseract-ocr postgresql-client postgresql-common libpq-dev
```
frankhereford, line 9: having `psql` and friends around is helpful debugging this flow. 🐘

flows/vision-zero/make_kv_store.py
```diff
@@ -25,8 +25,23 @@
...
+    "EMS_S3_SOURCE_BUCKET": os.getenv("EMS_S3_SOURCE_BUCKET"),
+    "EMS_S3_ARCHIVE_BUCKET": os.getenv("EMS_S3_ARCHIVE_BUCKET"),
+    "EMS_S3_SOURCE_PREFIX": os.getenv("EMS_S3_SOURCE_PREFIX"),
+    "EMS_S3_ARCHIVE_PREFIX": os.getenv("EMS_S3_ARCHIVE_PREFIX"),
 }
 
 json = json.dumps(kv_store)
 
-key_value_uuid = set_key_value(key="Vision Zero", value=json)
+key_value_uuid = set_key_value(key="Vision Zero Development", value=json)
```
frankhereford, line 47: In our next prefect huddle, I think I want to propose that we reevaluate the benefit of a `production` branch. I am finding that I have a predictable set of micro-changes in the checkout that is in production, including things like setting this `key` parameter back to `Vision Zero`. Nothing urgent, but I'd like to get those floating changes into version control.

flows/vision-zero/prefect_agent_environment_variables.env_template
```diff
@@ -9,12 +9,36 @@ AWS_ACCESS_KEY_ID=
...
 OCR_DIAGRAM_TARGET_PATH=
-OCR_BATCH_SIZE=
+OCR_BATCH_SIZE=10
 OCR_SINGLE_CRASH=
+
+AFD_DB_USERNAME=
+AFD_DB_PASSWORD=
+AFD_DB_HOSTNAME=
+AFD_DB_PORT=5432
+AFD_DB_DATABASE=
```
frankhereford, line 36: I am inclined to be totally fine with defining the database connection parameters again here, even though it's logical to assume that they will always match `DB_HOST`, `DB_USER`, etc as used in different flow in this project. Environment variables / flow parameters have no cost or performance impact, and I think how much they are isolated or sharing configuration variables with related flows can be assessed on a case by case basis by the developer.

Charlie-Henry (approved):

flows/vision-zero/make_kv_store.py
mddilley: I totally support of this. I think that I missed the initial discussions around this topic so I'd love to hear more. 🙏

mddilley (approved):
Looks good - moving on to review the other piece of this (https://github.com/cityofaustin/atd-vz-data/pull/1095). 🚀

patrickm02L (approved):
Relying on the Dev reviews here. 🤘

flows/vision-zero/make_kv_store.py
chiaberry: my memory is hazy, but I thought we had loosely agreed that a production branch didn't fit our needs. Are you proposing that we do need a production branch?

chiaberry (approved):

flows/vision-zero/make_kv_store.py
frankhereford: @chiaberry - your memory is totally right, I have thought and expressed that we didn't need a `production` branch for the VZ deployments in the past. I'm not positive that I think we would benefit from one now, but I have encountered a situation where I have some files on the `atd-prefect` production checkout which are floating and I can't just commit them back into main. 

Everytime I look at this, I think I'm coming up with a new different opinion on how best to handle it, so really, I wanted to brainstorm with the group using this particular case as an example starting off point. 

It's also worth mentioning that the production `atd-prefect` checkout is complicated by needing to run the moped test environment maker which has not made it into `main` yet.

PR #39, opened 2022-09-01
Prefect-ifying ATD Knack Services

Charlie-Henry:
This Prefect flow was created with [10002](https://github.com/cityofaustin/atd-data-tech/issues/10002) and [7442](https://github.com/cityofaustin/atd-data-tech/issues/7442) in mind to add roadway markings contractor work orders to Knack-services to upload data from Knack to Postgres, AGOL, and Socrata.

This flow uses the docker image automatically created by [this repo](https://github.com/cityofaustin/atd-knack-services). When it is run, it will pull the latest docker image and then run a series of commands inside that container.

In theory, this flow should be able to be configured to run any of our atd-knack-services DAGs currently in airflow.

More in README...

### DIFF
--- a/flows/knack/knack_services.py
+++ b/flows/knack/knack_services.py
@@ -71,7 +71,7 @@ def pull_docker_image():
 # Get the envrioment variables for the given app
 @task(
     name="get_env_vars",
-    # task_run_name="get_env_vars:{app}",
+    task_run_name="get_env_vars:{app}",
     max_retries=1,
     timeout=timedelta(minutes=60),
     retry_delay=timedelta(minutes=5),
@@ -86,7 +86,7 @@ def get_env_vars(app):
 # Get the last date (string) the flow was succesful
 @task(
     name="get_last_exec_time",
-    # task_run_name="get_last_exec_time (app: {app}, container: {container})",
+    task_run_name="get_last_exec_time (app: {app}, container: {container})",
     max_retries=1,
     timeout=timedelta(minutes=60),
     retry_delay=timedelta(minutes=5),
@@ -120,7 +120,7 @@ def get_last_exec_time(app, container, replace_data):
 # Records to postgrest
 @task(
     name="records_to_postgrest",
-    # task_run_name="records_to_postgrest (app: {app_name}, contianer: {container}, date_filter:{date_filter})",
+    task_run_name="records_to_postgrest (app: {app_name}, contianer: {container}, date_filter:{date_filter})",
     max_retries=1,
     timeout=timedelta(minutes=60),
     retry_delay=timedelta(minutes=5),
@@ -149,7 +149,7 @@ def records_to_postgrest(app_name, container, date_filter, environment_variables
 # Records to AGOL
 @task(
     name="records_to_agol",
-    # task_run_name="records_to_agol (app: {app_name}, container: {container}, date_filter: {date_filter})",
+    task_run_name="records_to_agol (app: {app_name}, container: {container}, date_filter: {date_filter})",
     max_retries=1,
     timeout=timedelta(minutes=60),
     retry_delay=timedelta(minutes=5),
@@ -178,7 +178,7 @@ def records_to_agol(app_name, container, date_filter, environment_variables):
 # Records to Socrata
 @task(
     name="records_to_socrata",
-    # task_run_name="records_to_socrata (app: {app_name}, container: {container}, date_filter: {date_filter})",
+    task_run_name="records_to_socrata (app: {app_name}, container: {container}, date_filter: {date_filter})",
     max_retries=1,
     timeout=timedelta(minutes=60),
     retry_delay=timedelta(minutes=5),
@@ -207,7 +207,7 @@ def records_to_socrata(app_name, container, date_filter, environment_variables):
 # Building AGOL segment geometries
 @task(
     name="agol_build_markings_segment_geometries",
-    # task_run_name="agol_build_markings_segment_geometries (layer: {layer}, date_filter: {date_filter})",
+    task_run_name="agol_build_markings_segment_geometries (layer: {layer}, date_filter: {date_filter})",
     max_retries=1,
     timeout=timedelta(minutes=60),
     retry_delay=timedelta(minutes=5),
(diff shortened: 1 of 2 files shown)


flows/knack/knack_services.py
```diff
@@ -0,0 +1,400 @@
...
+CONTAINER = "view_3628"
+LAYER_NAME = "markings_contractor_work_orders"
+# Parameter that will overwrite all data (ignores date)
+REPLACE_DATA = False
+APP_NAME_DEST = ""
+SOCRATA_FLAG = True
+AGOL_FLAG = True
+
+# Define current environment
+current_environment = "prod"
```
frankhereford, line 45: I think this variable is just used once, and it's used to build the name of the knack services. Maybe just define that end result variable here .. the idea being to more easily reveal how this string will get used? I thought changing this value would effect more behaviors and what-have-you about what the flow would do based on its name.

flows/knack/knack_services.py
```diff
@@ -0,0 +1,400 @@
...
+
+# Set up slack fail handler
+handler = slack_notifier(only_states=[Failed])
```
frankhereford, line 48: 🙏 I need to get better at notifications. Thanks for this example.

flows/knack/knack_services.py
```diff
@@ -0,0 +1,400 @@
...
+
+
+# Records to postgrest
+@task(
+    name="records_to_postgrest",
+    task_run_name="records_to_postgrest (app: {app_name}, contianer: {container}, date_filter:{date_filter})",
+    max_retries=1,
+    timeout=timedelta(minutes=60),
+    retry_delay=timedelta(minutes=5),
+    # state_handlers=[handler],
```
frankhereford, line 144: I see this line commented out. Is this something you can maybe remove everywhere before you merge, or comment back in?

flows/knack/knack_services.py
```diff
@@ -0,0 +1,400 @@
...
+        )
+        .decode("utf-8")
+    )
+    logger.info(response)
+    return response
+
+
+# Update the date stored in the key value in Prefect
+# if the upstream tasks were successful
+@task(trigger=all_successful)
```
frankhereford, line 286: TIL! This is a new way to add nodes to a DAG that I didn't know about - thank you.

flows/knack/knack_services.py
```diff
@@ -0,0 +1,400 @@
...
+    with case(to_knack, True):
+        knack_res = records_to_knack(
+            app_name,
+            container,
+            date_filter,
+            app_name_dest,
+            environment_variables,
+            upstream_tasks=[docker_pull, postgrest_res, to_knack],
+        )
+
```
frankhereford, line 385: I wonder if it would be helpful to `merge()` all of these `case()`es results back into a single node on the DAG and make it an upstream node of `update_last_exec_time()`?

frankhereford (approved):
It looks good! I put a few comments in the code with some TILs and some thoughts / questions. I don't see anything that should hold up merging though, but they would be nice to see addressed. Thanks for this!

flows/knack/knack_services.py
chiaberry: feels like a holdover from when we were going to have production flows and stagings flows running. The flow can just be named `knack_services_production` instead of using this variable too. 

Unless theres a chance we will have staging flows for knack services?

flows/knack/knack_services.py
```diff
@@ -0,0 +1,400 @@
...
+# Set up slack fail handler
+handler = slack_notifier(only_states=[Failed])
+
+# Logger instance
+logger = prefect.context.get("logger")
+
+# Select the appropriate tag for the Docker Image
+docker_env = "production"
+docker_image = f"atddocker/atd-knack-services:{docker_env}"
```
chiaberry, line 55: same here, would the `docker_env` ever not be production?

flows/knack/knack_services.py
```diff
@@ -0,0 +1,400 @@
...
+    # Key is unique based on the container and app
+    key = f"{app}:{container}-prev-exec"
+
+    # Set that key to the new date
+    prev_execs[key] = new_date
+    set_key_value(key="atd_knack_services_prev_exec", value=prev_execs)
+
+
+# Next, we define the flow (equivalent to a DAG).
+# Notice we use the label "test" to match this flow to an agent.
```
chiaberry, line 303: 🧹 from the template

flows/knack/knack_services.py
```diff
@@ -0,0 +1,400 @@
...
+    )
+    replace_data = Parameter("Replace all Data", default=REPLACE_DATA, required=True)
+    app_name_dest = Parameter(
+        "Records to Knack: App Name Destination", default=APP_NAME_DEST, required=False
+    )
+    soda_flag = Parameter("Send data to Socrata", default=True, required=True)
+    agol_flag = Parameter("Send data to AGOL", default=True, required=True)
+
+    # Based on provided parameters, skip or run some conditional tasks
+    build_geom, to_knack = determine_task_runs(layer, app_name_dest)
```
chiaberry, line 331: this is cool!

chiaberry (approved):
so sorry its taken me this long to review the PR

I really like this pattern of using parameters to tell the flow what to do. I left comments about removing the current_environment type variables, but now im thinking that a future way to handle running a flow with a staging env would be to use parameters??

flows/knack/knack_services.py
Charlie-Henry: Thank you, I'll zap these #'s to make sure we get slack notifications

flows/knack/knack_services.py
Charlie-Henry: I'm going to go ahead and zap this `current_environment` variable and just change the name to `knack_services_production`. I like @chiaberry's suggestion of using parameters for testing new code without having to go and re-register your flow with a different `docker_env`.

flows/knack/knack_services.py
Charlie-Henry: I haven't use `merge()` before, also sorry for responding super late on this comment @frankhereford. I think merge all of these `upstream_tasks` into one task? I'm not even sure I need `upstream_tasks` here anyway since they are now supplied to our task arguments now.

flows/knack/knack_services.py
Charlie-Henry: Refactored this into a parameter for our flow

frankhereford (changes requested):
Charlie, thanks for continuing to push this forward! This new work looks awesome. 

Since the reviews on it are a few months old and you've extended the content of this PR to include a new flow along with various other changes, do you mind re-requesting reviews when it's ready to merge?

Thanks!

Charlie-Henry:
Whoops yes, @frankhereford sorry for the noise on this. Basically the parameterized version of this flow was neat but it ended up being a bit of hassle to set up, and it was only being stored in the UI which if we lost that we'd have to manually go back and reset all of our parameters. I couldn't find a way to register a flow with parameters in the UI without just setting new defaults, which essentially just made a whole new file. So, now I'm following more of the pattern we had in Airflow where each flow will have its own file in this repo.

Charlie-Henry:
Going to close this and just open a new PR due to the shift in the direction of this knack services flow

PR #42, opened 2022-11-30
Submodule commit for moped-etl changes

chiaberry:
This is the companion prefect PR for the changes done in the atd-moped submodule https://github.com/cityofaustin/atd-moped/pull/887

I am wondering though, if I once that PR is approved and merged to the md_moped_test_create_flow branch, if then this branch needs to be updated with a new commit hash?

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,6 +1,8 @@
 # Directories
 .idea
 venv
+.DS_Store
+flows/.DS_Store
 
 # Files
 *.md5sum


johnclary (approved):
i'm not really sure what i'm approving but looks 💯 🚢 🚢 🚢 🚢 🚢 🚢 🚢

mddilley (approved):
Looks great to me! I think that you are right that you will need to update the submodule after merging the `atd-moped` branch that pairs with this one. 👍

PR #44, opened 2022-12-07
Create VZ Backup to S3 ETL

roseeichelmann:
This is still a WIP, I'm looking for any glaringly obvious issues with the script. I'll post some of the errors I've been getting as well, I'm just trying to get the get_backup() command to run at all before I worry about piping it and then running the get_aws_credentials() command. Once everything actually works I'll prefectify it.

### DIFF
--- a/flows/vision-zero/.gitignore
+++ b/flows/vision-zero/.gitignore
@@ -1,1 +1,2 @@
 *env
+atd_vz_backup/
--- a/flows/vision-zero/database_backup/database_backup.py
+++ b/flows/vision-zero/database_backup/database_backup.py
@@ -1,6 +1,16 @@
 import os
-import subprocess
 from datetime import datetime
+import docker
+from dotenv import load_dotenv, find_dotenv
+from pathlib import Path
+
+load_dotenv()
+
+env_path = Path('.')/'postgresql.env'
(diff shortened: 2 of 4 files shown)


roseeichelmann:
I'm running the script again and its been running for over an hour so I don't have any errors to show for it yet, @frankhereford could you remind me of that flag you were telling me about to only get back one table?

frankhereford:
@roseeichelmann wrote:

> I'm running the script again and its been running for over an hour so I don't have any errors to show for it yet, @frankhereford could you remind me of that flag you were telling me about to only get back one table?

Especially over the VPN, but even at the office, an hour is a "reasonable" expected time for this process, but yea, it's almost unworkable in the development cycle. 

Here is [the manual page](https://www.postgresql.org/docs/current/app-pgdump.html) for `pg_dump`, and in particular, I believe you may be seeking a combination of:
* `-t / --table=pattern` which says "only include this table(s)"
* `-T / --exclude-table=pattern` which says omit this table (thinking of the `atd_txdot_change_log` table)
* `-s / --schema-only` which omits all data and only pulls down the shape of the tables. This is probably the least helpful, but can be handy when experimenting with `pg_dump` on the command line.

PR #46, opened 2022-12-13
Pinning Prefect Version

Charlie-Henry:
We made this patch in main branch, but not in production so the auto-registering Github action was failing because pip was installing Prefect 2.0. Pinning here to 1.X. 

Only affects flows that get auto-registered so they must contain `if __name__ == "__main__":`


johnclary (approved):
🙌 🚢

PR #49, opened 2022-12-16
Service Bot Branch Name

Charlie-Henry:
Forgot to update this branch name in these flows 🤦 

Also note: `access_token_secret="GITHUB_ACCESS_TOKEN"` as we were getting rate limited as we were using a lot of github storage. This references the [Prefect Secrets](https://cloud.prefect.io/team/secrets).

### DIFF
--- a/flows/service-bot/intake_issues.py
+++ b/flows/service-bot/intake_issues.py
@@ -16,8 +16,6 @@ from datetime import timedelta
 # Prefect
 from prefect import Flow, task, Parameter
 from prefect.storage import GitHub
-from prefect.schedules import Schedule
-from prefect.schedules.clocks import CronClock
 from prefect.run_configs import LocalRun
 from prefect.engine.state import Failed
 from prefect.backend import get_key_value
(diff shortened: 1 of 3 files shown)


frankhereford (changes requested):
I think we need to test this configuration to make sure the GitHub auth is working as intended. It looks like you're passing the literal string `GITHUB_ACCESS_TOKEN` (and not the interpolated value held in a variable of that name) into the `access_token_secret` `GitHub()` configuration parameter. Am I reading this wrong?

Do you have a testing mechanism setup locally? I can set one up on one of the on-prem servers or my mac, but I don't want to duplicate any effort if I can avoid it. 

Thanks!

Charlie-Henry:
@frankhereford this tripped me up too but the `access_token_secret` is actually supposed to be a [string with they key of our secret we are referencing](https://docs-v1.prefect.io/api/latest/storage.html#github). Our secrets are stored [here](https://cloud.prefect.io/team/secrets). After I made this change last Friday, I haven't seen this error pop back up.

chiaberry:
Thanks prefect for being weird about variable names :)

frankhereford (approved):
Thanks @Charlie-Henry! I wouldn't have guessed that was the way it worked in a million years - sorry for the misunderstanding and error on my part. 🙏🏻

chiaberry:
It tripped me up too, do you think a comment or note for future us that says "yes this is supposed to be a string with the key of our secret we are referencing" would help?

Charlie-Henry:
@frankhereford no problem, it definitely _looks_ so wrong! I'll add that @chiaberry

chiaberry (approved):

mddilley (approved):
🚢

PR #50, opened 2022-12-29
Vision Zero DB / RR Permission Maintainer

frankhereford:
# DB Permission Maintainer

### 🔧 Action
This PR intends to close [#11074](https://github.com/cityofaustin/atd-data-tech/issues/11074). 
This concept is derived from [#10251](https://github.com/cityofaustin/atd-data-tech/issues/10251).

### 🧹 Description
This is a bare-bones flow which consumes some environment variables which define a postgres endpoint, a list of staff members, and a list of schemas. The flow will then (re-)grant permission to select on the tables within these imports for these staff members.

This task is required for all tables which are created which happens from time-to-time in our dev process and daily for the `import` schema in the VZDB.

### 🥼 Testing
Inside the dockerized agent container, the script is located here: `/root//root/db_permission_grantor/db_permission_grantor.py`. 

The script can be tested by:

* Updating your environment file,
* Building the VZ dockerized agent,
* Starting the container with `docker compose up -d`,
* `docker exec -it prefect-agent-vz bash` to get a shell on the container,
* Flipping the boolean on line 27,
* And then executing the program. 

* The logging should indicate the SQL `GRANT` statements issued.

This is a long way to go for a pretty simple script. I would be happy to demo its operation or we may be satisfied with a code review - up to you!

Thanks!


flows/vision-zero/db_permission_grantor/db_permission_grantor.py
```diff
@@ -0,0 +1,70 @@
...
+    DB_NAME = os.getenv("DB_NAME")
+    SCHEMAS_CDL = os.getenv("DB_PERMISSION_SCHEMAS")
+    STAFF_CDL = os.getenv("DB_PERMISSION_STAFF")
+else:
+    DB_HOST = kv_dictionary["DB_HOST"]
+    DB_USER = kv_dictionary["DB_USER"]
+    DB_PASS = kv_dictionary["DB_PASS"]
+    DB_NAME = kv_dictionary["DB_NAME"]
+    SCHEMAS_CDL = kv_dictionary["DB_PERMISSION_SCHEMAS"]
+    STAFF_CDL = kv_dictionary["DB_PERMISSION_STAFF"]
```
johnclary, line 35: @frankhereford would it be possible to fetch the list of staff users through a DB query instead of having to maintain the list of staff in the kv store?

johnclary (approved):
this passes the eyeball test, with the caveat that this needs a README here or in gitbook. thank you @frankhereford!

flows/vision-zero/db_permission_grantor/db_permission_grantor.py
frankhereford: @johnclary In a way, yes, but we'd still need to keep track of a list of 'non-user' accounts, which may be better.

Here's an example of `SELECT rolname from pg_roles;`with other staff members names blurred out.
roles

We could hard-code in all the non-staff members and do some set operations to get at a dynamic list of staff members. I'm not sure if this really helps us all that much - maybe I'm not seeing the whole thing clearly.

Same, but to lesser degree, with the schemas - I think it's generally safer to have an explicit list of what we're targeting for our `GRANT`s, as we're giving privilege. I'd be more inclined to have a wider net if we were taking away access. 

I'm super open to discussion - what do you think?

frankhereford:
@johnclary wrote:
> <snip> this needs a README here </snip>

Good call! Putting one together right now - thanks for this reminder!

patrickm02L (approved):
None shall pass without the appropriate GRANT statements!


https://user-images.githubusercontent.com/5566727/209981869-b277bd5a-e5df-4ccc-b95b-2b88052b0ce2.mp4

flows/vision-zero/db_permission_grantor/db_permission_grantor.py
```diff
@@ -0,0 +1,70 @@
...
+kv_dictionary = json.loads(kv_store)
+
+DB_HOST = None
+DB_USER = None
+DB_PASS = None
+DB_NAME = None
+SCHEMAS_CDL = None
+STAFF_CDL = None
+
+if False:
```
chiaberry, line 22: I have vague memory of this pattern from before, but can't remember why we use `if False`. is it for local development?

frankhereford: It's an ugly pattern, and it's entirely from my hands. 🙋🏼‍♂️

I really want to write a function which detects if it's running in 'agent' mode or if it's being run via an invocation of `.run()` in the script, and based on that pulls variables from the KV store or the actual environment.

I've lamented not knowing how to detect running in either state, so i've just been propagating this ugly boolean to switch between the development/production configurations. I also have not put much of an attempt to make that distinction - it would be a neat micro-skill-share in a prefect meeting discovery maybe.

frankhereford:
@johnclary @chiaberry 👋 I've re-requested a review. I've implemented John's suggestion to use a shared role `staff` to which each staff member is granted membership in. It's this `staff` role which is re-granted permissions by the ETL.

Side note: The staff role and membership in it are already setup on the VZDB server. Roles are managed at the database level, so it's not possible to configure it in staging and then later configure the production database -- they come as one.

Thanks again for the suggestion, and I hope the re-request isn't too much noise. Let me know if I can answer any questions!

flows/vision-zero/db_permission_grantor/db_permission_grantor.py
frankhereford: I did a quick search in the prefect python module documentation, and I didn't see a method that was jumping out at me that would indicate if the flow is running via a schedule & agent vs. being invoked on the command line. If you think it's valuable, we can spin off an issue to work out this development vs. production configuration mechanism.

Charlie-Henry: Maybe you could do something like [this](https://github.com/cityofaustin/atd-prefect/blob/74eec77e8cb04ce6ce8015e39304014708e82e77/flows/test/skillshare.py#L17) where you have a env variable defined inside your dockerized agent that would tell you if you are running locally?

mddilley (approved):
Awesome - very cool to see this making its way into the VZ stack. 🙏 I know this is working great where it lives 💪 , but I had a couple small issues getting the stack up locally. 

Do we have any docs about getting this up and running from scratch? The two main things were errors coming from copying ssh config files in `/prefect_bootstrap/` that rightfully shouldn't be in the repo (lines 12-16 in the Dockerfile), and then I didn't have a Prefect API key in my .env so the container kept erroring and restarting. Figured that one out by running in attached mode. Definitely not a huge priority atm, and I don't want to block your progress on the LDM. 👍

PR #52, opened 2023-01-13
Parking Data Secrets Refresh

Charlie-Henry:
After changing some secrets earlier this week, this flow was failing since I did not have a task to refresh our secrets each time we run this flow. This will also be needed for the ODP migration next Tuesday. Didn't realize that you can use `flow.chain` with arguments for your flows, makes your code much easier to read!

I also increased the timeout for some tasks that pull data from Flowbird since we had a significant backlog of data to download and they have a whooping one minute cooldown on API requests.

Note to self before merge:
- [x] change `current_envrionment` to `production`
- [x] change branch reference to `main`

After merge verify that the flow's schedule is turned on and `test` parking flow is turned off.

### DIFF
--- a/flows/parking/atd_parking_data_meters_txn_history.py
+++ b/flows/parking/atd_parking_data_meters_txn_history.py
@@ -27,7 +27,7 @@ from prefect.tasks.docker import PullImage
 from prefect.utilities.notifications import slack_notifier
 
 # Define current environment
-current_environment = "test"
+current_environment = "production"
 
 # Set up slack fail handler
 handler = slack_notifier(only_states=[Failed])
@@ -229,7 +229,7 @@ with Flow(
     storage=GitHub(
         repo="cityofaustin/atd-prefect",
         path="flows/parking/atd_parking_data_meters_txn_history.py",
-        ref="ch-parking-data-secrets",  # The branch name, staging is not being used here
+        ref="main",  # The branch name, staging is not being used here
     ),
     # Run config will always need the current_environment
     # plus whatever labels you need to attach to this flow
--- a/flows/parking/atd_parking_data_meters_txn_history.py
+++ b/flows/parking/atd_parking_data_meters_txn_history.py
@@ -103,7 +103,7 @@ def get_env_vars():
 @task(
     name="parking_transaction_history_to_s3",
     max_retries=1,
-    timeout=timedelta(minutes=60),
+    timeout=timedelta(minutes=180),
     retry_delay=timedelta(minutes=5),
     state_handlers=[handler],
     log_stdout=True,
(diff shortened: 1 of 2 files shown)


flows/parking/atd_parking_data_meters_txn_history.py
```diff
@@ -88,16 +88,27 @@ def pull_docker_image():
...
 
+# Get the envrioment variables
+@task(
+    name="get_env_vars",
+    state_handlers=[handler],
+    log_stdout=True,
+)
+def get_env_vars():
+    environment_variables = get_key_value(key=f"atd_parking_data_meters")
+    return environment_variables
```
frankhereford, line 99: 👍 🚀 

Just wanted to note this pattern that Charlie has here. It's really helpful, and I'd suggest we follow it for prefect flows going forward unless there's a compelling reason to for a given flow.

flows/parking/atd_parking_data_meters_txn_history.py
```diff
@@ -217,20 +229,21 @@ def update_last_exec_time():
     storage=GitHub(
         repo="cityofaustin/atd-prefect",
         path="flows/parking/atd_parking_data_meters_txn_history.py",
-        ref="production",  # The branch name, staging is not being used here
+        ref="ch-parking-data-secrets",  # The branch name, staging is not being used here
```
frankhereford, line 232: These branch names in here are probably not ideal, and I like your idea that you shared with me recently about using a value out of the environment / kv store to control this sort of thing. I'm not suggesting that you adopt it here unless you want to and have the bandwidth, but it occured to me that your solution would fit in with this. Am I fitting this all together right in my head?

flows/parking/parking_data_reconciliation.py
```diff
@@ -397,26 +409,27 @@ def update_last_exec_time():
...
-        ref="production",  # The branch name
+        ref="ch-parking-data-secrets",  # The branch name
     ),
     # Run config will always need the current_environment
     # plus whatever labels you need to attach to this flow
     run_config=UniversalRun(labels=[current_environment, "atd-data02"]),
     schedule=Schedule(clocks=[CronClock("00 5 * * *")]),
 ) as flow:
+    environment_variables = get_env_vars()
     flow.chain(
```
frankhereford, line 420: TIL `flow.chain()` - that's swank.

frankhereford (approved):
shipit

flows/parking/atd_parking_data_meters_txn_history.py
Charlie-Henry: Yeah we should pair on trying to set something up where we can check what branch we're in and just plug that in here? Maybe `production` agents just use `main` and otherwise it's up to you to manually set the branch reference. This gets me all the time!

Charlie-Henry:
Merging before ODP migration changes later today

PR #56, opened 2023-02-03
Remove protection of certain columns in the CRIS import script

### LINKED ISSUE #10652: Populate `cris` schema with data from CRIS
### Acceptance Criteria
* [x] Branch and modify the CRIS import script / flow found in [cris_import.py](https://github.com/cityofaustin/atd-prefect/blob/main/flows/vision-zero/cris_import/cris_import.py) to populate the `cris` schema
  * This should remove a fair chunk of code related to "important' field handling and change conflict resolution
* [x] `vision-zero` script commands to get data into the database, similar to the "Recreate Environment Path" section [here](https://github.com/cityofaustin/atd-moped/pull/792)
  * [x] Turn down `graphql-engine`
  * [ ] ~Disconnect other PG connections if possible~ 
    * This is unreliable at best; I suggest we rely on the devs to disconnect their SQL clients when populating the DB
  * [x] Pipe SQL files into `psql`
  * [x] Turn up `graphql-engine`

frankhereford:
This PR intends to close https://github.com/cityofaustin/atd-data-tech/issues/10652. 

NB: this PR is into `moped-project` which is a feature branch we're currently running out of for our atd-data03 production deployment. I will rebase the PR to merge into main if needed.

Testing:

* In your VZ local stack, checkout the `layer-data-model` branch if needed
* Spin up a local DB and populate it with data
* Create the `cris` schema, as found here: [create_cris_schema.sql](https://github.com/cityofaustin/atd-vz-data/blob/import-cris-data/atd-toolbox/materials_from_ldm_conversion/create_cris_schema.sql)
* Back in this repo, spin up a container of the VZ dockerized-agent / execution environment and attach to it
* Use the CRIS extract account to cause CRIS to push an export of data for some time interval of your choosing to the SFTP endpoint
* When the file is available, either copy it to your local machine and adjust the tasks used in the flow to pick it up directly (at or around [line 485 in the cris_import.py script](https://github.com/cityofaustin/atd-prefect/blob/import-cris-data-ldm/flows/vision-zero/cris_import/cris_import.py#L484-L489)), or let the flow take care of pulling it down from the SFTP site on its own. It can be useful to keep a copy somewhere on the SFTP endpoint so you can put it back to rerun the flow a second time if you need to.
* Run the flow and observe that it will now create updates when processing the CRIS extract and does so with no consideration as to what is an "important" or editable field in the VZ. 
  * There should be no way for an update from CRIS to cause the VZ conflict resolution system to be used.


frankhereford:
Y'all, I just realized that I missed a totally important part of the acceptance criteria on this PR, switching the storage schema for data out of the `public` schema and into the `cris` one. I'm going to remove my requests for review while I fix that - sorry for the noise.

frankhereford:
I have created and applied a new label to this PR, `no merge`. I believe we're going to get `moped-project` into `main` in the near future, and I would like to wait until that happens so that I can branch a new collector branch off of that named `layer-data-model`. I will rebase this PR into that collector branch so that we can accumulate layer data model changes and not block unrelated updates to prefect while it's being worked on. 

Please feel free to continue with reviews if you'd like -- this is more a reminder to myself to not let this this merge until the dust settles on the repo a little.

flows/vision-zero/cris_import/cris_import.py
```diff
@@ -373,9 +409,6 @@ def align_records(typed_token, dry_run):
         # Query the list of columns in the target table
         target_columns = util.get_target_columns(pg, output_map, table)
 
-        # Get the list of columns which are designated to to be protected from updates
-        no_override_columns = mappings.no_override_columns()[output_map[table]]
-
```
mddilley, line 378: 👋

flows/vision-zero/cris_import/cris_import.py
```diff
@@ -512,7 +505,7 @@ def align_records(typed_token, dry_run):
     align_records_token = align_records(typed_token=typed_token, dry_run=dry_run)
 
     # push up the archives to s3 for archival
-    uploaded_archives_csvs = upload_csv_files_to_s3.map(extracted_archives)
+    # uploaded_archives_csvs = upload_csv_files_to_s3.map(extracted_archives)
```
mddilley, line 508: Since I'm out of the loop, is this a temporary change or are we stopping the S3 archives?

flows/vision-zero/cris_import/cris_import.py
```diff
@@ -272,7 +301,14 @@ def pgloader_csvs_into_database(directory):
...
 def remove_trailing_carriage_returns(data_loaded_token):
 
-    pg = psycopg2.connect(host=DB_HOST, user=DB_USER, password=DB_PASS, dbname=DB_NAME, sslmode="require", sslrootcert="/root/rds-combined-ca-bundle.pem")
+    pg = psycopg2.connect(
+        host=DB_HOST,
+        user=DB_USER,
+        password=DB_PASS,
+        dbname=DB_NAME,
+        sslmode=DB_SSL_REQUIREMENT,
+        sslrootcert="/root/rds-combined-ca-bundle.pem",
```
mddilley, line 310: Nice - this will make troubleshooting a bit easy if we run into any SSL issues in the future. Also, do we need to pass this cert? 

Last week, I noticed that our password manager entry called **Vision Zero Database Configuration** has a note that says
> SSL is required to connect to this service, but should not need a special certificate installed.

I forgot to ask about this then so I figured now would be a good time to check if we need to update that.

flows/vision-zero/cris_import/lib/sql.py
```diff
@@ -40,36 +39,31 @@ def get_column_operators(
             
             # there are two normal ways to be equal. Either be of the same value and type or /both/ be undefined.
             comparison_clause = f"""
-                (  (public.{output_map[table]}.{column['column_name']} = {DB_IMPORT_SCHEMA}.{table}.{column['column_name']})
-                OR (public.{output_map[table]}.{column['column_name']} IS NULL AND {DB_IMPORT_SCHEMA}.{table}.{column['column_name']} IS NULL)
+                (  (cris.{output_map[table]}.{column['column_name']} = {DB_IMPORT_SCHEMA}.{table}.{column['column_name']})
```
mddilley, line 42: 🚀  Excited about this schema coming to life 🌱

flows/vision-zero/prefect_agent_environment_variables.env_template
```diff
@@ -44,3 +44,4 @@ EMS_S3_SOURCE_PREFIX=
 EMS_S3_SOURCE_PREFIX=
 
 DB_PERMISSION_SCHEMAS="import, hdb_catalog, public"
+DB_SSL_REQUIREMENT=required
```
mddilley, line 47: I think this needs to be updated to `require` 🔒

mddilley (approved):
I tested this by copying a CRIS extract from the SFTP to my local machine and then updating the flow to point to my local DB and load up its fresh `cris` schema. The proof is in the screenshot 📸  - it's alive!!! I had a couple questions, and there is one small change needed to the `env` template, but all looks good! 🚀 👏 

Screenshot 2023-02-21 at 1 26 36 PM

mddilley:
Almost forgot that I ran into [this line](https://github.com/cityofaustin/atd-prefect/blob/3e5028b33198c18edda3a23cb9f042670cdf7ee6/flows/vision-zero/cris_import/cris_import.py#L125) when I was running this locally. It will throw an undefined error if `rsync` ever fails `false` => `False`. I should have caught that the first time I reviewed this code a while back! 🐍

flows/vision-zero/cris_import/cris_import.py
frankhereford: Thanks Mike for catching that. I pushed up a change that reenables the task and restores its inter-task dependencies to get it back into the right part of the flow. 🙏🏻

flows/vision-zero/prefect_agent_environment_variables.env_template
frankhereford: Great catch, thank you! I didn't realize this was broken because I had adjusted it in my actual local env file.

I have settled on setting the suggested default for this to be 'prefer.' The idea is that when running locally, SSL/TLS isn't a concern, and further, the postGIS image we use (nor the one provided by the official postGIS folks) has support for SSL/TLS out of the box. It certainly can be made to support it, but we don't have a good use case for going through the trouble of providing legit certificates to make that happen. 

`prefer` is chosen because if someone does fire this up to communicate over the internet, hopefully the server will support SSL/TLS and the script will use it, but if not, such as when using this as a local service for development etc, it can fail to established an encrypted channel and still work. 

For our production deployment on `atd-data03`, we should stick with require, I think.

PR #58, opened 2023-02-09
ATD-Finance-Data to Prefect 2

Charlie-Henry:
Converting an existing Prefect 1.0 flow to Prefect 2.

Instructions to build and deploy this flow on data-03:

Build Deployment yaml file: 
> `$ prefect deployment build flows/finance-data/finance_data_to_s3.py:main --name "Finance Data Publishing" --cron "13 7 * * *" --pool atd-data-03 -q default -sb github/finance-data-wip -o "deployments/finance_data_to_s3.yaml" `

Breaking this down, the above command creates a yaml file with: 
- `name` is the name of our flow
- `pool` is the work pool of our agent
- `cron` is the schedule of our flow (UTC)
- `q` is the name of our agent
- `sb` is the storage block of our flow (you'll need to create one if you are working on a feature branch)
- `o` is the name of yaml file and where we're storing it. Created a deployments subdirectory to store these files.

Then, apply this deployment
> `$ prefect deployment apply deployments/finance_data_to_s3.yaml`

I'm storing our secrets as a JSON block in Prefect, probably less than ideal if we want to store it somewhere not in plaintext?

I've used the [slack notifications automation](https://app.prefect.cloud/account/62be21ad-87d4-4858-bc47-a49e1fd29df7/workspace/4e909dab-3495-4e26-9403-9b633af2c970/automations/automation/ddd5aae8-cdc6-4368-a96a-b790ec49079d) @chiaberry set up. We can edit this to add the flows we want slack messages for.


flows/finance-data/.prefectignore
```diff
@@ -0,0 +1,41 @@
+# prefect artifacts
```
johnclary, line 1: > By default, Prefect uploads all files in the current folder to the configured storage location (local by default) when you build a deployment.
> 
> If you want to omit certain files or directories from your deployments, add a .prefectignore file to the root directory. .prefectignore enables users to omit certain files or directories from their deployments.
> 
> Similar to other .ignore files, the syntax supports pattern matching, so an entry of *.pyc will ensure all .pyc files are ignored by the deployment call when uploading to remote storage.

Interesting. I wonder if this can live in the root of the repo and then we don't need to add one to every directory? Also, the prefect docs are not clear, but since we're using github storage i don't think the deployment is actually uploading files anywhere, right?

flows/finance-data/finance_data_to_s3.py
```diff
@@ -15,23 +18,9 @@
...
-from prefect.backend import set_key_value, get_key_value
-from prefect.triggers import all_successful
-from prefect.tasks.control_flow import merge
-
-from prefect.utilities.notifications import slack_notifier
+from prefect import flow, task, get_run_logger
+from prefect.blocks.system import JSON
 
-# Set up slack fail handler
-handler = slack_notifier(only_states=[Failed])
```
johnclary, line 31: can you add the `SlackWebhook` for notifications?

deployments/finance_data_to_s3.yaml
```diff
@@ -0,0 +1,55 @@
+###
+### A complete description of a Prefect Deployment for flow 'Finance Data Publishing'
+###
+name: Finance Data Publishing
+description: null
```
johnclary, line 5: Can you add a description of what this does and the path to the finance data repo?

deployments/finance_data_to_s3.yaml
```diff
@@ -0,0 +1,55 @@
...
+  _is_anonymous: true
+  block_type_slug: process
+  _block_type_slug: process
+storage:
+  repository: https://github.com/cityofaustin/atd-prefect
+  reference: ch-prefect-2-test
+  access_token: null
+  include_git_objects: true
+  _block_document_id: 386bb209-603f-42ab-b361-d07010704f11
+  _block_document_name: finance-data-wip
```
johnclary, line 43: i'm curious to hear what y'all think about managing production vs development flow data. i'm mostly worried about accumulating cruft in our deployments and and storage blocks.

johnclary (changes requested):
thank you for doing this!!

flows/finance-data/finance_data_to_s3.py
chiaberry: There's an Automation in the cloud UI to trigger a slack notification upon flow failure, its configurable to select "all flows" or just certain flows. Charlie set up slack notifications via the UI. 

We could also use the SlackWebhook if we wanted to further refine the error messages or send messages on other states/results.

flows/finance-data/.prefectignore
chiaberry: I'm leaning towards .gitignoring the .prefectignore. I'll put it on the agenda for Tuesday

deployments/finance_data_to_s3.yaml
chiaberry: I'm thinking that whoever creates a storage block used in testing flows, should then delete that block after the flows are approved. (Or reuse the block, but in any case, not keep accumulating blocks)

flows/finance-data/.prefectignore
Charlie-Henry: @johnclary that's correct, maybe that's something can configure but I doubt we want Prefect auto-committing stuff to github. Currently, I get a warning saying it's not configured to upload to our storage location. I'm fine with gitignoring it!

deployments/finance_data_to_s3.yaml
Charlie-Henry: It does feel crufty to me too 😢 

Another potential solution would be to make an `atd-prefect-dev` or something S3 bucket where it will upload and store our code and use that as our catch-all storage block for all things that are not yet merged into `main`.

flows/atd-finance-data/finance_data_to_s3.py
```diff
@@ -0,0 +1,188 @@
...
+             then moves it along to Knack and socrata.
+
+Build Deployment yaml file:
+$ prefect deployment build flows/atd-finance-data/finance_data_to_s3.py:main \
+--name "Finance Data Publishing" \
+--cron "13 7 * * *" \
+--pool atd-data-03 -q default \
+-sb github/finance-data-wip \
+-o "deployments/finance_data_to_s3.yaml" \
+--description "Repo: https://github.com/cityofaustin/atd-finance-data, Gets Finance data from a database, places it in an S3 bucket, then moves it along to Knack and socrata."
```
Charlie-Henry, line 15: Pointing out in the latest Prefect release they added a `--description` CLI argument!

mddilley:
@Charlie-Henry I'm going to continue looking at this in the morning while working on https://github.com/cityofaustin/atd-data-tech/issues/11418 so that I get a better feel for it all. I don't have a good enough grasp to give valuable feedback quite yet - especially about John's question about development vs. prod. I'll be back with a review!

chiaberry (approved):

deployments/finance_data_to_s3.yaml
frankhereford: I've just spent a minute fussing with the prefect 2 REST api, "documented" [here](https://app.prefect.cloud/api/docs). 

I've been able to pull a list of active blocks associated to our account & workspace, and more generally automate interacting with the prefect service. What would you all think about the idea of writing a prefect ETL to clean up blocks that we don't put on a "save-forever" list after, say, 60 days?

An automatic clean-up robot as it were?

.gitignore
```diff
@@ -5,3 +5,5 @@ venv
 # Files
 *.md5sum
 *.env
+
+.prefectignore
```
frankhereford, line 9: Not a big deal, but if you don't mind, can you throw a line break at the end of this last line? No worries if not.

frankhereford (approved):
Get out your captain's hat and send this one out to sea, because it's ready to ship. Thanks Charlie!

Charlie-Henry:
@frankhereford I like your idea about using the REST API to drop our unused blocks. 

One thing I had an idea for but couldn't wrap my head around the "documentation" was to set up a flow that will remove late service-bot's intake issues flows that run every three minutes. If we have our agent go down, those will pile up in the queue and run all at once when it restarted and it starts to create dupe issues in Github. It looks like this is possible with the Prefect "automations" but we only get to use three at our free-tier workspace and we already have two configured.

johnclary:
🤯 ugh, i forgot about the failed flow runs piling up. Airflow has the `catchup` [flag](https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/dag-run.html#catchup) which works perfectly for this. it's frustrating b/c i don't think we have a single ETL that would benefit from having missed flow runs stay in the queue.

frankhereford:
@johnclary wrote:
> 🤯 ugh, i forgot about the failed flow runs piling up. Airflow has the `catchup` [flag](https://airflow.apache.org/docs/apache-airflow/stable/core-concepts/dag-run.html#catchup) which works perfectly for this. it's frustrating b/c i don't think we have a single ETL that would benefit from having missed flow runs stay in the queue.

We actually got burned by this problem pretty solidly recently after the latest linux upgrades from CTM. I forgot to start an agent on atd-data02, and when I did later that day, it fired up the Knack → GitHub ETL script for our new issue creation. I was able to kill it before it created to many duplicate entries, but not before it got the load on atd-data02 up in the 300s. I was surprised that we were able to recover the machine without a hard reset TBH. 

The prefect 1 UI offers a functionality labeled as "clear late flows" which was the step we needed; it empties out all queued jobs that should have been picked up in the past. I'm not sure where to find this in prefect 2, but I'm very hopeful it's in there somewhere.

deployments/finance_data_to_s3.yaml
mddilley: Those are cool ideas, Frank and Charlie!  I mentioned this in dev sync today, but I was wondering if there would be a downside to creating both production and staging/development blocks per flow. That way someone working on a flow could update the reference in the staging block and go from there. I might have overlooked something, though. Most of our development on flows seems to be one person working on one flow at a time but that could be a limitation.

I've been thinking about this since the service bot flows have config for test/dev/staging and production. I created staging and production secret blocks to see how that goes.

flows/atd-finance-data/finance_data_to_s3.py
```diff
@@ -0,0 +1,188 @@
...
+
+# Docker settings
+docker_env = "production"
+docker_image = "atddocker/atd-finance-data"
+
+
+@task(
+    name="get_env_vars",
+    retries=10,
+    retry_delay_seconds=timedelta(seconds=15).seconds,
```
mddilley, line 40: I'm split on whether I should use `timedelta` vs. using an integer for the seconds in the script that I'm working on. Just thinking out loud here!

Also, I am curious about the timeout vs. retry and if there is a reason for the update. I was trying to make the same decision earlier today on the service bot intake script, and I could use your thoughts on this for sure.

flows/atd-finance-data/finance_data_to_s3.py
```diff
@@ -0,0 +1,188 @@
...
+
+        master_agreements_res = upload_to_s3(
+            environment_variables.value["data-tracker"], "master_agreements"
+        )
+        logger.info(master_agreements_res)
+
+        fdus_res = upload_to_s3(environment_variables.value["data-tracker"], "fdus")
+        logger.info(fdus_res)
+
+        if all([task_orders_res, units_res, fdus_res]):
```
mddilley, line 183: Awesome - TIL that `.all` is the way to do this in Python. 🐍

mddilley (approved):
Looks awesome, Charlie! Thanks for migrating this. I reviewed the code but didn't test the script. This one was also really helpful in working on the service bot intake flow since it also pulls a Docker image. I'll stick to the same kind of updates that you made here so that we have some consistency from flow to flow. 🚢 👏

mddilley:
Forgot to comment IRT the late flows piling up in the queue - I'm trying [what I described here](https://austininnovation.slack.com/archives/CHZE6BC6L/p1678745191242269) out for the service bot flows. Queues look like a good way to organize the runs that each work pool is handling and also gives us more control over the number of concurrent runs per queue. I set the service bot queue limit to 1 since it defaults to unlimited.

roseeichelmann (approved):
Prefect 2.0 here we come 🚢🫡 I didn't test just reviewed the code but it looks great, thanks for working on these migrations. Having y'alls scripts as examples will be super helpful when it comes time for me to write my first Prefect 2 script

Charlie-Henry:
@johnclary I believed I addressed your requested changes?

PR #59, opened 2023-02-13
Auth to docker hub when starting the dockerized agent which runs a docker agent

frankhereford:
## TLDR
Tell the prefect docker-agent to auth to docker when it starts up.

## Background

`atd-data03` runs a dockerized agent environment which is configured to start a prefect docker agent. This agent came out of a sync @Charlie-Henry and I had maybe 6 months ago, give or take. He expressed interest in having a docker-agent in the prefect sense of the word, and he's been using it successfully to fire up flows bundled in docker images since.

This sounds like a painfully inception-like configuration, but the dockerized-agent does offer some ease of configuration perks, and it does pass down the host's docker socket fill into the running containers, so this really just works. The container's docker-agent starts docker containers (from within the dockerized-agent env) which actually run as siblings to the subject dockerized agent on the host. They each think they are starting children, but they all share the same docker socket from the host, so they all actually just run as siblings directly on `atd-data03` with no performance loss.

The dockerized agent's place to put bootstrapping actions for container start-up ends up making this change somewhat easier, and avoids having to bundle the actual secret into the image. I did add those two environment variables to environment variable file, which you don't see in this PR, but it's the secrets file, so that's expected. You do see where they are reflected in the template file, which was missing.


chiaberry (approved):

johnclary:
+1 thank you @frankhereford!! 🚢

PR #60, opened 2023-02-13
Remediate CRIS import failure

mddilley:
Associated issues
https://github.com/cityofaustin/atd-data-tech/issues/11402

This adds `sslmode` to the pgloader commands run by the CRIS import flow. Still kind of a mystery about why this worked prior to February 7 since the switch to `pgloader` was made before then and the flow ran successfully. I also added a log message that happens if the `pgloader` command fails to successfully exit. Need some feedback on that since I'm not sure that will surface in the Prefect Cloud UI. 🙏 

I found https://github.com/dimitri/pgloader/issues/768#issuecomment-693390290 which led me to try `sslmode=allow` to get his back up and running. `require` is what we use with `psycopg2` but that errored out with `pgloader`. [Here are the PG docs about the SSL modes](https://www.postgresql.org/docs/current/libpq-ssl.html) but "allow" translates to "use SSL if it is there" as opposed to "require" which translates to "must use it." I've proved that our RDS will not accept non-SSL connections with both `psql` and trying that mode with `pgloader`. I think that makes it safe to go ahead with this mode set.

Let me know if anyone disagrees or if there is something that I'm overlooking. Really enjoyed learning more on this one, and these is so much more.

Testing

The best way that I can think of to test this is to use the test fixture that Frank and I used to test this. That file can be removed after this PR is approved.

1. SSH into `atd-data03`
2. `docker exec -it prefect-agent-vz bash` to attach to the VZ dockerized agent and start a bash shell
3. cd to `/root/cris_import` and then edit `cris_test_fixture.py` to revert the changes made in this PR
4. Save and then run `cris_test_fixture.py` - it should fail and log `KABOOM!` and then the DB connection errors as it maps through invocations of `pgloader_csvs_into_database()`
5. Change it back and see that there are no errors 

Ship list
- [ ] Remove `cris_test_fixture.py` and `test_db_connection.py` from `/root/cris_import` in the dockerized agent
- [ ] Code reviewed
- [ ] Product manager approved


frankhereford:
Upon acceptance and merger of this PR, the following task should be executed as soon as is reasonably possible: https://github.com/cityofaustin/atd-data-tech/issues/11463.

flows/vision-zero/cris_import/cris_import.py
```diff
@@ -244,13 +243,14 @@ def pgloader_csvs_into_database(directory):
...
-                CONNECTION_STRING = f'postgresql://{DB_USER}:{DB_PASS}@{DB_HOST}:5432/{DB_NAME}'
+                # See https://github.com/dimitri/pgloader/issues/768#issuecomment-693390290
+                CONNECTION_STRING = f'postgresql://{DB_USER}:{DB_PASS}@{DB_HOST}:5432/{DB_NAME}?sslmode=allow'
 
                 with open(command_file, 'w') as file:
                     file.write(f"""
 LOAD CSV
     FROM '{directory}/{filename}' ({headers_line})
-    INTO  {CONNECTION_STRING}?import.{table} ({headers_line})
+    INTO  {CONNECTION_STRING}&import.{table} ({headers_line})
```
frankhereford, line 253: For future reviewers - this PR ended up with a lot of extransious commits from our current branching configuration in `atd-prefect`. The above lines are the fix which he came up with. 

The two commits of actual interest in this PR are:
* https://github.com/cityofaustin/atd-prefect/pull/60/commits/99334e6ac79dd1dbb7c2b45a3238844f5452acc8
* https://github.com/cityofaustin/atd-prefect/pull/60/commits/99334e6ac79dd1dbb7c2b45a3238844f5452acc8

mddilley:
Since this originally branch from `atd-data03-dockerized-agent-deployment`, I created https://github.com/cityofaustin/atd-prefect/pull/61 to clean up the diff. I'll close this one out and ask for reviews in the other.

PR #62, opened 2023-02-22
Migrate the hr-knack-banner script Prefect 2

chiaberry:
Addresses https://github.com/cityofaustin/atd-data-tech/issues/11354

I updated the flow to 2.0, also deleted the email part of the flow since that was not ever used. 

to test: I don't have a staging environment set up for this ETL, so if you run it you will be updating the prod knack app. 

1. start a new virtual env, install requirements
2. `python flows/atd_knack_banner/knack_banner.py` will run the flow. 

I have an issue open to figure how to make the deployments happen when a branch is merged to main,  but for now, writing the deployment yaml file happens on the command line, I used this to create it
`prefect deployment build flows/knack/knack_banner.py:knack_hr_banner_flow --cron "45 13 * * *" -q atd-data-03 --name "HR Knack Banner" -o "deployments/knack-banner.yaml" -sb github/atd-prefect-main-branch --apply` 
Because the storage block (`-sb`) points to the main branch, the scheduled flow won't be successful until this is merged to main.


docs/prefect-1.0/01-basics.md
```diff
@@ -46,7 +46,7 @@ $ virtualenv venv
 $ source venv/bin/activate
 
 # 3. Install the requirements for local development
-$ pip install -r requirements.txt
+$ pip install -r requirements-legacy.txt
```
chiaberry, line 49: I moved all the old docs to a different folder

flows/knack/knack_banner.py
```diff
@@ -4,124 +4,58 @@
...
-) as send_email_flow:
-    get_data_flow_run_id = create_flow_run(flow_name=get_data_flow.name)
-    script_result = get_task_run_result(
-        get_data_flow_run_id, task_slug="knack-banner-copy"
-    )
-    formatted_data = format_email_body(script_result)
-    send_email_flow.chain(script_result, formatted_data, email_task(msg=formatted_data))
+@flow(name=f"Knack HR Banner")
+def knack_hr_banner_flow():
+    logger = get_run_logger()
```
Charlie-Henry, line 55: Doesn't look like you are using this logger. I was sad to learning in my testing that Prefect 2 doesn't allow us to define loggers as a global so you either have to pass this logger around or get a new one in each task like what you have

deployments/knack-banner.yaml
```diff
@@ -0,0 +1,56 @@
+###
```
Charlie-Henry, line 1: Think this should be `knack_banner.yaml` with the underscore instead of dash if we want to be consistent with the deployment yaml file matching flow file name.

flows/knack/knack_banner.py
```diff
@@ -4,124 +4,58 @@
 Name: Knack Banner HR App integration
 Description: Update knack HR app based on records in Banner and CTM
 Schedule: "45 13 * * *"
-Labels: atd-data02, knack
 """
```
Charlie-Henry, line 7: Maybe something for prefect meeting but do we want to include in the docstring how to deployment build and deployment apply for this flow file? 

[I've been doing it](https://github.com/cityofaustin/atd-prefect/blob/c121eaa09f23ea05410e41a6d544ddad335c16be/flows/finance-data/finance_data_to_s3.py#L9) because I kept losing my command and having to retype it all the time.

Charlie-Henry (approved):

deployments/knack-banner.yaml
chiaberry: thank you i will fix this

flows/knack/knack_banner.py
chiaberry: you're right, i didnt understand why the global logger wasnt working and this was left over from when i thought i was going to pass it instead of creating a new one in that task

flows/knack/knack_banner.py
chiaberry: I really like this idea

deployments/knack_banner.yaml
```diff
@@ -0,0 +1,56 @@
...
+  _is_anonymous: true
+  block_type_slug: process
+  _block_type_slug: process
+storage:
+  repository: https://github.com/cityofaustin/atd-prefect/
+  reference: main
+  access_token: null
+  include_git_objects: false
+  _block_document_id: d0cc8e56-412b-49b3-bb9c-33b522c64736
+  _block_document_name: atd-prefect-main-branch
```
johnclary, line 44: i'm worried about the management of all these blocks becoming an overwhelming mess.
Screen Shot 2023-02-23 at 10 59 15 AM

requirements.txt
```diff
@@ -1,55 +1,108 @@
+aiosqlite==0.18.0
```
johnclary, line 1: its best to only list the packages you pip installed, and not include those packages' dependencies. If we want to pin the the minor version of prefect, that'd be

```
prefect==2.8.*
```

johnclary (changes requested):
thank you for making this happen!!! i left questions and comments.

johnclary:
one more thought on file organization. i think flows should live in folders that match their repo name. so this flow would be in `/atd-knack-banner` and charlie's would be in `atd-finance-data`. i know there might be edge cases where a flow runs code from two repos, but in general i think we should strive to have it be as clear as possible the relationship b/t a flow and the source code it runs.

chiaberry:
@johnclary I like the idea of organizing the flows by repo name. an improvement over our airflow repo's one folder named "dags". I'll bring it up at the prefect meeting on tuesday.

flows/knack/knack_banner.py
```diff
@@ -4,124 +4,62 @@
...
-from prefect.tasks.prefect import create_flow_run, get_task_run_result
-from prefect.storage import GitHub
-from prefect.run_configs import UniversalRun
-from prefect.engine.state import Failed, TriggerFailed, Retrying
-from prefect.schedules import Schedule
-from prefect.schedules.clocks import CronClock
-from prefect.backend import get_key_value
-from prefect.triggers import all_successful
+from prefect import flow, task, get_run_logger
+from prefect.blocks.notifications import SlackWebhook
```
johnclary, line 19: i don't think this is being called anywhere?

deployments/knack_banner.yaml
```diff
@@ -0,0 +1,56 @@
+###
+### A complete description of a Prefect Deployment for flow 'Knack HR Banner'
+###
+name: HR Knack Banner
+description: null
```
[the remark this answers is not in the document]
chiaberry: I didn't catch that the auto generated yaml file would include "### A complete description" even if we didn't pass in the description as one of the configs. 

This is making me lean more towards us devs having to create our own deployment files from the CLI and then a github action could be the step that applies them.

deployments/knack_banner.yaml
Charlie-Henry: We could potentially offload our secret storage to S3 or our password manager of choice?

Charlie-Henry:
One thing I just remembered is that because we have `if __name__ == "__main__":` in our Prefect 2 flows, the github actions will try to register this to Prefect 1.0. Not that it'll work or that it's a big deal, but to turn those off do we have to open a PR to delete the workflows?

chiaberry:
@Charlie-Henry good catch about the github workflow. I am going to disable it for now.

chiaberry:
This should be ready for re-review. 

@johnclary let me know what other changes you want addressed

mddilley:
Dropped [this note](https://github.com/cityofaustin/atd-prefect/pull/58#issuecomment-1459058344) in Charlie's PR. Same applies here - I want to give y'all feedback. I know these PRs have been hanging around for a while!

deployments/knack_banner.yaml
```diff
@@ -0,0 +1,57 @@
+###
+### A complete description of a Prefect Deployment for flow 'Knack HR Banner'
+###
+name: HR Knack Banner
+description: Gets updates from HR Banner endpoint and CTM email list and updates Knack
```
Charlie-Henry, line 5: New update yesterday allows for markdown in the[ description section of your flow!](https://www.loom.com/share/e41ab81d4b954829ab55cd08f207ecb6) 😮

deployments/knack_banner.yaml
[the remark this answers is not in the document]
johnclary: interesting—that makes sense to me!

deployments/knack_banner.yaml
frankhereford: Is there a known limit on the length of the secret's value in a secret block? I don't see anything jumping out at me from the docs, so I'm inclined to want to use those to house our secrets as they are obfuscated in the prefect UI. 

I think it would be pretty clean to have a single secret per flow, and have that secret be a JSON blob with key-value pairs as needed, or even a more complex shape if the flow warrants it. 

I'm not against using S3, but I'd want to warn against it because that pushes a lot of responsibility on us to make sure that resource is properly secured from the public (or even non-DTS folks) yet remains available to flows as needed. We'd have to have a secret to give us access to our other secrets, which doesn't feel awesome.

I think it would be really slick if we could use 1Password as the secret store (it's like its true raison d'etre, if anything), but I'd also want to see it's reliability proven up with a discovery day experiment or something along those lines.

frankhereford (approved):
This worked well. I included a few questions and comments which are probably way late to the game, but certainly nothing to keep this from getting merged. Thanks Chia!

requirements.txt
```diff
@@ -1,55 +1 @@
...
-tblib==1.7.0
-text-unidecode==1.3
-toml==0.10.2
-toolz==0.11.2
-tornado==6.1
-urllib3==1.26.7
-websocket-client==1.2.1
-wrapt==1.13.3
-zict==2.0.0
+prefect==2.8.*
```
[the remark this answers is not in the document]
chiaberry: This flow (and I think the majority of our flows will eventually, considering so many of them will be atd-knack-services) fires up a docker container, I am envisioning the requirements.txt file just to be to run prefect locally that kicks off the docker container housing the actual script. Besides the vision zero flows, which flows would need their own packages? Or are you thinking hypothetically?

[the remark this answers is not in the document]
Charlie-Henry: Yeah I think the pattern we have here is fine. One `requirements.txt` file that will serve as the default for our agent. The best way to handle repo-specific requirements would be in docker similar to this flow, but if you need to supply it directly to Prefect you can do it using the Docker Infrastructure block and supply a `EXTRA_PIP_PACKAGES` field. I think there's a way to create blocks in the CLI so we could just document that in a README if we want to document the extra packages we're installing.  

[This tutorial is pretty good at explaining how to use docker blocks as "infrastructure" in Prefect.](https://towardsdatascience.com/create-robust-data-pipelines-with-prefect-docker-and-github-12b231ca6ed2)

deployments/knack_banner.yaml
mddilley: I like the idea of a central secrets store like 1Password like Frank mentioned. If that doesn't pan out, we could always looks into AWS Secrets Manager as a central source of secrets. Some of these flows could add up to a lot of API calls over time and get $$$, though. 

I'm trying out the Secret blocks for the atd-service-bot flows to store an encoded JSON string and then load it in the script. One downside that I see to this is that we'll probably need to document the expected fields in the JSON or it might be hard to modify/create since you can't retrieve the value once it is set in the Prefect UI (which is a good thing for security).

flows/atd-knack-banner/knack_banner.py
```diff
@@ -0,0 +1,64 @@
...
+    # Environment Variables stored in JSON block in Prefect
+    return JSON.load("atd-knack-banner")
+
+
+# Retrieve the provider's data
+@task(
+    name="update HR employees in knack",
+    retries=2,
+    retry_delay_seconds=300,
+    task_run_name="atd-knack-banner/update-employees",
```
mddilley, line 35: Does this do anything besides giving it a name? I had a hard time wrapping my head around this parameter when I was reading through the docs.

mddilley (approved):
Thanks for migrating this over to 2.0! 🚀 This PR was super helpful in getting the right changes into the flow that I'm working on. I did not fire this up locally, but all of the code looks great! 🙌

roseeichelmann (approved):
i reviewed the code and it looks good, thanks for helping to lead the way with these migrations. also ty for updating the README! its been interesting reading the convos about secret blocks vs other ways to store the secrets and while i dont have much input i like the different ideas im hearing! i definitely need to catch up with some prefect 2 tutorials

johnclary (approved):
🚢🚢🚢🚢 🙌

flows/atd-knack-banner/knack_banner.py
chiaberry: i dont think it does, I thought it would show up in the UI, but i think i named everything too similarly so I can't tell

PR #63, opened 2023-03-09
Update VZ ETLs to use a bastion host to connect to the VZDB

frankhereford:
## Intent
This PR will allow https://github.com/cityofaustin/atd-data-tech/issues/11677 to be closed.
This PR has a sibling PR which needs to be merged alongside this one. You can find it [here](https://github.com/cityofaustin/atd-vz-data/pull/1187/files). 

## Method
Because we're trying to disallow users from connecting to the production VZDB from the COA network, but yet, still allow our ETLs to connect -- and since both originate from the same IP at the edge of the COA network -- the bastion host is needed to act as a gatekeeper. 

This PR updates the VZ ETLs to use a SSH tunnel via the bastion to connect to the database.

## Testing
This is a testing bear. Luckily, each script is idempotent, so they can be run multiple times against our production DB. Also, due to the nature of the change, if the `ssh` tunnel fails to get instantiated, there is no way for the script to make changes. The CRIS import script requires a file be available to import on the SFTP server. 

I have tested the method used here when developing the `db_permission_granter.py` and it worked as expected.

## The Ask
I have PM'd with @chiaberry, and she said she'd support an alternative "testing" method here. I'd like to ask you all to review the code but don't feel like you need to get it running. With your approvals, I will install these on production and make sure they each work, tweak anything needed, and then with that, I will complete the merge. When doing this, I'll de-schedule the ETLs for a morning, and kick them off individually, and when everything is working as intended, I'll re-enable that schedule.


mddilley (approved):
Wow - never really thought about programatically tunneling like this! You flew through this update - thank you! 🔐 I read through the code and it makes sense to me so it gets an enthusiastic ✅ 🚀

chiaberry (approved):

johnclary (approved):
I have stared at the code. Thank you for this quick work, @frankhereford!

🚢

PR #64, opened 2023-03-15
Upgrade service bot intake issue flow to Prefect 2

mddilley:
https://github.com/cityofaustin/atd-data-tech/issues/11418

This PR migrates the service bot intake (Knack service request to Github issue) flow to Prefect 2. Going to branch from here for the other two flows to keep this from getting too big and hard to test.

I tried out Secret blocks and also attempted staging and production runs/environments. Prod is default but you can pass `env` (new parameter) and `docker_tag` (existing parameter) to use the staging Secret block and test an updated image if needed. Super open to feedback on that since not all flows will have a clear staging env like this one since it has a staging/test Knack app.

Concurrency limit is set to 1 as [mentioned here](https://austininnovation.slack.com/archives/CHZE6BC6L/p1678745191242269) - still need to discuss how to clear the queue of late flows in the next Prefect meeting

### Testing

1. Submit a service request in the Knack app called `DTS | Data and Technology Services Portal | PRODUCTION Copy` which I've linked below but before that:
         - Be sure to name the request something obvious like "TEST Service Bot issue" because it will create a real issue [here](https://github.com/cityofaustin/atd-data-tech/issues)
         - Be sure to check that the url matches the one below and that you are not creating a new service request in the production app or else it will get picked up by the production bot instead of this Prefect 2 one (sometimes clicking links in Knack apps can redirect you to the production app without you realizing it)
         - https://atd.knack.com/dts--data-and-technology-services-portal--production-copy#new-service-request/ 
         - Also be aware that this will end up triggering an email that goes to the transportation.data inbox which could be confusing to the service desk primary. I've given Andrew S. a heads up about testing this week.

3. Go to the Prefect 2 cloud https://app.prefect.cloud/ and sign in with transportation.data@austintexas.gov which sends a magic link. After you are signed in go to https://app.prefect.cloud/account/62be21ad-87d4-4858-bc47-a49e1fd29df7/workspace/4e909dab-3495-4e26-9403-9b633af2c970/deployments/deployment/4e3e7729-43bf-46da-a0ee-8fed2d11e9cd

4. In the top right, click the **Run** button and choose **Custom run** from the dropdown
5. Double check the Parameters at the bottom of this view show `staging` as the value for `env` and then click **Run**
6. You can watch the logs to make sure the flow succeeds and then go to https://github.com/cityofaustin/atd-data-tech/issues
7. You should see your service request in the form of a new Github issue which you can close
8. Clean out the emails that this generated in transportation.data inbox


deployments/atd_service_bot_intake.yaml
```diff
@@ -0,0 +1,61 @@
...
+# The work queue that will handle this deployment's runs
+work_queue_name: atd-service-bot
+work_pool_name: atd-data-03
+tags: []
+parameters: {}
+schedule:
+  cron: '*/3 * * * *'
+  timezone: null
+  day_or: true
+is_schedule_active: false
```
mddilley, line 17: I switched this to off because I didn't want the flow to immediately schedule when I applied the deployment.

flows/atd-service-bot/intake_issues.py
```diff
@@ -0,0 +1,93 @@
...
+
+
+@flow(name="Service Bot: Intake Issues")
+def intake(docker_tag="production", env="production"):
+    """Intakes new issues from Knack to Github
+
+    Keyword arguments:
+    docker_tag -- the docker tag to use (default "production")
+    env -- the environment to use (default "production")
+    """
```
mddilley, line 81: I parameterized the env so that I could test this out with the staging secret block easily. Also, the docstring is a little repetitive with the notes that we have at the beginning of these files so I can drop it if needed.

deployments/atd_service_bot_intake.yaml
```diff
@@ -0,0 +1,61 @@
...
+  _block_type_slug: process
+
+###
+### DO NOT EDIT BELOW THIS LINE
+###
+flow_name: 'Service Bot: Intake Issues'
+manifest_path: null
+storage:
+  repository: https://github.com/cityofaustin/atd-prefect
+  reference: md-11418-service-bot-intake
```
mddilley, line 37: I thought of this reference line as something that we could update in the UI to point new code (see `  _block_document_name: atd-service-bot-staging` a few lines down). My branch name will still get committed here, though. Not really sure what the right answer is here. 🤔

deployments/atd_service_bot_intake.yaml
```diff
@@ -0,0 +1,61 @@
...
+  title: Parameters
+  type: object
+  properties:
+    docker_tag:
+      title: docker_tag
+      default: production
+      position: 0
+    env:
+      title: env
+      default: production
```
Charlie-Henry, line 57: Whoa this is new to me, didn't realize there's two places to put parameters in the deployment yaml file. TIL! It wasn't auto populating the parameters for me in my deployment file, so it's nice you figured it out. 

It looks like it auto populates because you set them as defaults in the main method
```
def intake(docker_tag="production", env="production"):
```

where I was passing them as args but not setting default...
```
def main(arg1, arg2)
```

Charlie-Henry (approved):
Thank you Mike! I tested using the staging environment and it works 💯 

My comment about the github access token needs to be taken care of (I think @frankhereford can make one?) before we switch this on but other than that 🚢

deployments/atd_service_bot_intake.yaml
johnclary: i'm having a hard time keeping these params straight, but my initial reaction to storage is that we shouldn't ever need to reference a non-production or non-staging storage block in a deployment. especially when the ETL itself is dockerized. that is making the not-great assumption that flows will just work when we deploy them on our remote hosts.  and i'm probably forgetting a bunch of other considerations.

anyway, i know we need to revisit our prefect CI in general so i don't feel too strongly about this.

flows/atd-service-bot/intake_issues.py
```diff
@@ -0,0 +1,93 @@
...
+    return docker_image
+
+
+# Get the envrioment variables based on the given environment
+@task(name="get_env_vars", timeout_seconds=60)
+def get_env_vars(env):
+    logger = get_run_logger()
+
+    # Environment Variables stored in secret block in Prefect
+    secret_block = Secret.load(f"atd-service-bot-{env}")
```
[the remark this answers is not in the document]
johnclary: ugh, i wish i had helpful thoughts on the best may to manage secrets + env vars without having, e.g., our Socrata key, stored in a dozen different places 🤔

roseeichelmann (approved):
tested this with the staging environment and it works great! also cleaned out the emails and issue this generated. thank you for working on these! 🙏✨

deployments/atd_service_bot_intake.yaml
chiaberry: I'm still trying to decide if I just dont apply the deployment until after its merged? I think thats how I'm leaning. but not strongly.

flows/atd-service-bot/intake_issues.py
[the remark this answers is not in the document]
chiaberry: I am working on an atd-knack-services prefect flow and also thinking about how I don't like having the same keys or ids stored in a ton of different blocks. 
Maybe one secret that is the socrata key, multiple secret blocks that are reused.

chiaberry (approved):
didn't test but reviewed the code -- thank you! if its better to keep track of your work by creating new issues for the other aspects of the service bot work, by all means feel free.

flows/atd-service-bot/intake_issues.py
```diff
@@ -0,0 +1,93 @@
+"""
+Name: ATD Service Bot: Intake Issues
+Description: Sends new issue data from our service portal (knack) to our Github
+Schedule: */3 * * * * (AKA once every 3 minutes)
+Work queue concurrency limit: 1
+prefect deployment build flows/atd-service-bot/intake_issues.py:intake \
+    --cron "*/3 * * * *" --pool atd-data-03 -q atd-service-bot \
+    --name "Service Bot: Intake Issues" -o "deployments/atd_service_bot_intake.yaml" \
+    -sb github/atd-service-bot-staging --skip-upload \
```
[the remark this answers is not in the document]
mddilley: Awesome - thanks for the heads up on this! I added the access token to the storage block

deployments/atd_service_bot_intake.yaml
mddilley: Yep, that is what I'm wondering too. I was thinking about having staging deployments and production deployments. A staging deployment would be the starting point for the next developer to make changes and test around. 

Maybe, then the CI could look at staging deployment files on merge to `main`, and, if there was a change, apply the production deployment which always points at `main`. Definitely something that we could talk about in the meeting tomorrow.

mddilley:
I updated the folders and deployment yaml names and applied them as we talked about the other day in dev sync. We'll see how it goes, and I'm ready to change it up if we need to. Thanks for the reviews, y'all!

PR #65, opened 2023-03-20
Upgrade service bot Socrata and update flows to Prefect 2

mddilley:
https://github.com/cityofaustin/atd-data-tech/issues/11418

This picks up where https://github.com/cityofaustin/atd-prefect/pull/64 left off and updates the following flows to Prefect 2:
- **Issues to Socrata**: takes Github and Zenhub data and pushes issues to the open dataset
- **Update Knack issues**: keeps the Knack records in sync with updates that happen in Github and Zenhub to issues with the **Project Index** label

### Testing

These are a bit tough to test so I added steps so y'all can confirm that the flows did their job when I ran them.

#### Issues to Socrata flow
1. Go to the test service issues open dataset at https://datahub.austintexas.gov/Transportation-and-Mobility/TEST-ATD-Data-Tech-Services-Issues/93ru-6p6e/
2. Login with the password manager entry called **Socrata Admin | City Data Hub/EDP**
3. If you sort descending by `created_at` then you should see issues in the dataset from today (March 21st) and that means that the script worked 🎉

#### Update Issues flow
1. Go to https://builder.knack.com/atd/dts--data-and-technology-services-portal--production-copy/records/objects/object_30?builder_table_search=service+bot and login using either your personal login (if you have access to this app) or the Knack builder one in the password manager
2. You should see a row in the `project_evaluation` table that was created after I briefly added the **Project Index** tag to https://github.com/cityofaustin/atd-data-tech/issues/11740 and ran this flow which looks to update **Project Index** labeled issues. The flow caught the update in the title from `🤖 THIS IS A SERVICE BOT TEST ISSUE` to `🤖 THIS IS A SERVICE BOT TEST ISSUE BEING UPDATED` and updated this row.


flows/atd-service-bot/helpers.py
mddilley, line 1: I added this helper module since these three flows share these tasks. Maybe this could be even more generalized, but this keeps these three flows grouped together and portable.

chiaberry: I like it!

chiaberry (approved):
Thanks for the clear steps on how to see that the script works. 

After you get approvals, are you going to update the deployment to point to the correct storage block? Still trying to figure out how to ensure the deployments match what we are going to save as the deployment files in github.

mddilley:
@chiaberry I'm still thinking about that, but my current idea is that I'll add a production GH block that points to `main` and then apply that after merge. The CI could take care of that eventually for us.

Charlie-Henry (approved):
👍  didn't run your test steps but looks good! Thanks Mike

roseeichelmann (approved):
followed your steps to see that the script worked, this looks good to me! i like the helpers module that you added. thanks for working on these mike 🚢

deployments/atd_service_bot_socrata.yaml
```diff
@@ -0,0 +1,62 @@
...
+
+###
+### DO NOT EDIT BELOW THIS LINE
+###
+flow_name: 'Service Bot: Issues to Socrata'
+manifest_path: null
+storage:
+  repository: https://github.com/cityofaustin/atd-prefect
+  reference: md-11418-service-bot-socrata
+  access_token: '**********'
```
frankhereford, line 39: I found the `atd-service-bot-staging` block which is where this access token appears to be stored in the prefect web UI. I find it interesting that it's represented in this file with a masked value instead of a boolean indicating that it simply exists, so that's what got me thinking about it. 

I don't see any problem with prefect having a value it can presumably pass down to a running flow to use to get files from GitHub over an auth'd connection, but I guess it seems different to me since the other flows that use a GH backed storage block don't have one. 

Does this need to be here because the script which is called by this flow need it to publish issues as `atdservicebot`? 

I'm sorry if this question is uniformed. I'm sure this is fine, and I am curious more than anything.

frankhereford (approved):
Thank you Mike, this looks great.  🚢 🚢 🙏

deployments/atd_service_bot_socrata.yaml
mddilley: Not at all! Just saw this but it relates to [this thread](https://austininnovation.slack.com/archives/CHZE6BC6L/p1670439656261679) and an issue that Charlie ran into in Prefect v1 too. In that case, this part happened in the code.

PR #66, opened 2023-03-21
Add flow for corridor retiming knack services

chiaberry:
The ETL side of https://github.com/cityofaustin/atd-data-tech/issues/11181

I took most of this work from Charlie's work on migrating atd-knack-services to prefect 2.

### DIFF
--- a/deployments/knack-services-corridor-retiming.yaml
+++ b/deployments/knack-services-corridor-retiming.yaml
@@ -2,9 +2,10 @@
 ### A complete description of a Prefect Deployment for flow 'Knack Services: Data Tracker Corridor Retiming'
 ###
 name: 'Knack Services: Corridor Retiming'
-description: Wrapper ETL for the atd-knack-services docker image 
-             with defined commands for the corridor retiming view in Data Tracker app
-version: 5efc182879c4d579a2c9c66926dc3af4
+description: Wrapper ETL for the https://github.com/cityofaustin/atd-knack-services
+  docker image with defined commands for the corridor retiming view in Data Tracker
+  app
+version: faff2b2e23e3884231c48f005b5e3c36
 # The work queue that will handle this deployment's runs
 work_queue_name: default
 work_pool_name: atd-data-03
@@ -59,4 +60,4 @@ parameter_openapi_schema:
   - commands
   - env_block_name
   definitions: null
-timestamp: '2023-03-29T00:31:14.258616+00:00'
+timestamp: '2023-03-29T02:08:50.804728+00:00'
--- a/deployments/knack-services-corridor-retiming.yaml
+++ b/deployments/knack-services-corridor-retiming.yaml
@@ -4,11 +4,12 @@
 name: 'Knack Services: Corridor Retiming'
 description: Wrapper ETL for the atd-knack-services docker image 
              with defined commands for the corridor retiming view in Data Tracker app
-version: 8234eb85e30f1ed0cc5bdaf51a0b8bb2
+version: 5efc182879c4d579a2c9c66926dc3af4
 # The work queue that will handle this deployment's runs
 work_queue_name: default
 work_pool_name: atd-data-03
-tags: []
+tags:
+- atd-knack-services
 parameters: {}
 schedule:
   cron: 45 11 * * *
@@ -58,4 +59,4 @@ parameter_openapi_schema:
   - commands
   - env_block_name
   definitions: null
-timestamp: '2023-03-21T16:28:24.077342+00:00'
+timestamp: '2023-03-29T00:31:14.258616+00:00'
--- a/deployments/knack-services-corridor-retiming.yaml
+++ b/deployments/knack-services-corridor-retiming.yaml
@@ -1,0 +1,61 @@
+###
+### A complete description of a Prefect Deployment for flow 'Knack Services: Data Tracker Corridor Retiming'
+###
+name: 'Knack Services: Corridor Retiming'
+description: Wrapper ETL for the atd-knack-services docker image 
+             with defined commands for the corridor retiming view in Data Tracker app
+version: 8234eb85e30f1ed0cc5bdaf51a0b8bb2
+# The work queue that will handle this deployment's runs
+work_queue_name: default
+work_pool_name: atd-data-03
+tags: []
+parameters: {}
+schedule:
+  cron: 45 11 * * *
+  timezone: null
+  day_or: true
+is_schedule_active: null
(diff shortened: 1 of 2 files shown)


flows/atd-knack-services/corridor-retiming.py
```diff
@@ -0,0 +1,145 @@
...
+@task(
+    name="update_exec_date",
+    retries=10,
+    retry_delay_seconds=timedelta(seconds=15).seconds,
+)
+def update_exec_date(env_block_name):
+    # Update our JSON block with the date of last successful flow execution
+    block = JSON.load(env_block_name)
+    block.value["PREV_EXEC"] = datetime.today().strftime("%Y-%m-%d")
+    block.save(name=env_block_name, overwrite=True)
```
frankhereford, line 109: This is game changing. I didn't realize we had such a readily available mechanism to save state in the prefect ecosystem. I can see a world where we track moped test deployments in a block and use it to clean up after us automatically after some fixed duration.

flows/atd-knack-services/corridor-retiming.py
```diff
@@ -0,0 +1,145 @@
...
+    if docker_res:
+        commands_res = docker_commands(environment_variables, commands, logger)
+    # After success, update prev_execution date in prefect json block
+    if commands_res:
+        update_exec_date(env_block_name)
+
+
+if __name__ == "__main__":
+    app_name = "data-tracker"  # Name of knack app
+    container = "view_3814"
```
frankhereford, line 133: I don't think this warrants the work to write it out, but I think this `container` variable is being assigned the Knack view number --  that could be pulled out of this prefect code and put in an environment variable or a argument to the process. The idea behind that is to keep that bright line between application specific code and the orchestration code.

mddilley (approved):
Thank you for bringing this one over to Prefect 2! I did not run this since I saw that `quirky-husky` 🤪 🐕 and others ran successfully. Everything looks awesome! ✨

deployments/knack-services-corridor-retiming.yaml
```diff
@@ -0,0 +1,61 @@
...
+###
+name: 'Knack Services: Corridor Retiming'
+description: Wrapper ETL for the atd-knack-services docker image 
+             with defined commands for the corridor retiming view in Data Tracker app
+version: 8234eb85e30f1ed0cc5bdaf51a0b8bb2
+# The work queue that will handle this deployment's runs
+work_queue_name: default
+work_pool_name: atd-data-03
+tags: []
+parameters: {}
```
Charlie-Henry, line 12: I think if you leave the `parameters` parameter blank it will not add default parameters to your flow. I have been putting them here manually but not 100% sure you have to.

deployments/knack-services-corridor-retiming.yaml
```diff
@@ -0,0 +1,61 @@
+###
+### A complete description of a Prefect Deployment for flow 'Knack Services: Data Tracker Corridor Retiming'
+###
+name: 'Knack Services: Corridor Retiming'
+description: Wrapper ETL for the atd-knack-services docker image
```
Charlie-Henry, line 5: Add link to repo in description, I think the description now supports markdown if you want to try that.

flows/atd-knack-services/corridor-retiming.py
Charlie-Henry: Maybe making a custom Prefect block for Knack would be a cool discovery day project for someone to embark on? 😉

flows/atd-knack-services/corridor-retiming.py
```diff
@@ -0,0 +1,145 @@
...
+
+    # Get env vars and pull the latest docker image
+    environment_variables = get_env_vars(env_block_name)
+    docker_res = pull_docker_image()
+
+    # Append appropriaste date argument to our commands list
+    commands = determine_date_args(environment_variables, commands)
+
+    if docker_res:
+        commands_res = docker_commands(environment_variables, commands, logger)
```
Charlie-Henry, line 125: Just a general comment to the team:
Maybe my pattern here is not ideal. Should we have a task that runs a single `command` and put our for loop in `main`?
Pro: We'd get a task for each docker command in the UI with better logging when something breaks
Con: Not sure if the tasks would then run asynchronously, this would likely break things here since we want to update postgres then upload that data to Socrata.

Charlie-Henry (approved):

flows/atd-knack-services/corridor-retiming.py
chiaberry: credit for this all goes to @Charlie-Henry

flows/atd-knack-services/corridor-retiming.py
roseeichelmann: this is an interesting idea, getting a task for each docker command to see what exactly breaks sounds like a huge plus

roseeichelmann (approved):
looks good :) 🚢

Charlie-Henry:
One thing I was just reminded of was how we were storing secrets in airflow. I started an [atd-knack-services block](https://app.prefect.cloud/account/62be21ad-87d4-4858-bc47-a49e1fd29df7/workspace/4e909dab-3495-4e26-9403-9b633af2c970/blocks/block/8fe5dfac-7cde-4f50-af99-380f71ab6bee) where we have each app name as a key and the separate config for each one. In theory, we could use just this one block for all knack service flows?

Like:
```
{
   "data-tracker": {env json},
   "smart-mobility": {env json},
   "finance-purchasing":{env json}
}

```

Only concerning thing was adding a key for the prev exec date, which artbox is stored just as an additional key `artbox_signals_prev_exec`? Maybe we could add a key like `PREV EXECS` with it's on json blob?

flows/atd-knack-services/corridor-retiming.py
```diff
@@ -0,0 +1,145 @@
...
+
+"""
+Name: ATD Knack Services: Corridor Retiming
+Description: Wrapper ETL for the atd-knack-services docker image 
+             with defined commands for the corridor retiming view in Data Tracker app
+Create Deployment:
+$ prefect deployment build flows/atd-knack-services/corridor-retiming.py:main \
+--name "Knack Services: Corridor Retiming" --pool atd-data-03 \
+--cron "45 11 * * *" -q default -sb github/atd-prefect-main-branch \
+-o "deployments/knack-services-corridor-retiming.yaml" --skip-upload
```
Charlie-Henry, line 11: I've added a tag to mine, might be useful
```
-tag atd-knack-services
```

PR #68, opened 2023-04-03
Parking Data Flow to Prefect 2

Charlie-Henry:
https://github.com/cityofaustin/atd-data-tech/issues/11831

Moves parking data flows to Prefect 2. 

Similar pattern as the previous PRs I've opened except in this flow we have  a `prev_month` argument that will collect data from the current month and the previous month around the beginning and end of each month. This should hopefully catch cases where data gets updated and we still have check for updates.

Reminder to update the deployment to the production github storage block when we are about to merge this.


flows/atd-parking-data/parking_data_processing.py
```diff
@@ -0,0 +1,226 @@
...
+    --name "atd-parking-data: Parking Data Processsing" \
+    --pool atd-data-03 \
+    --cron "35 8 * * *" \
+    -q default \
+    -sb github/ch-parking-data-prefect2 \
+    -o "deployments/parking_data_processing.yaml"\
+    --skip-upload \
+    --description "Repo: https://github.com/cityofaustin/atd-parking-data Scripts that download and process parking data for finance reporting."
+ 
+$ prefect deployment apply deployments/parking_data_processing.yaml
```
mddilley, line 19: Good call on adding the apply command here 👍

flows/atd-parking-data/parking_data_processing.py
```diff
@@ -0,0 +1,226 @@
...
+
+@task
+def add_command_arguments(commands, s3_env, start_date, prev_month):
+    # Adds additonal command arguments based on provided arguments
+    # and the previous exec date.
+    output_commands = []
+    for c in commands:
+        if "txn_history.py" in c:
+            c = f"{c} --env {s3_env}"
+            c = f"{c} --start {start_date}"
```
mddilley, line 115: This is a cool way to build up these commands!

mddilley (approved):
Looks great, Charlie! Thanks for making these updates, and thanks for explaining the goal of the previous month part of this code. Didn't test this because I know the best way to test some of these is to get them in service. 🚀 🚢

frankhereford (commented):
This looks great Charlie. Thanks for consolidating those two ETLs into this, and I agree with Mike, you've got a good thing going here where you're building up the arguments needed based on the input you receive. 🚢🚢🚢

PR #69, opened 2023-04-05
Signal Comms to Prefect 

Charlie-Henry:
Ports this [airflow ETL](https://github.com/cityofaustin/atd-airflow/blob/master/dags/atd_signal_comms.py) to Prefect. 

Part of: https://github.com/cityofaustin/atd-data-tech/issues/9471

Currently, data-03 is not whitelisted onto the signal network so none of the devices are currently able to be pinged.


Charlie-Henry:
now in airflow 2!

PR #70, opened 2023-04-12
Add flow to update traffic incidents

chiaberry:


flows/atd-traffic-incident-reports/traffic-incident-reports.py
```diff
@@ -0,0 +1,87 @@
...
+    client = docker.from_env()
+    client.images.pull(docker_image)
+    return True
+
+
+# Retrieve the provider's data
+@task(
+    name="update_postgrest_with_traffic_incidents",
+    retries=2,
+    retry_delay_seconds=300,
```
Charlie-Henry, line 58: Since the schedule is every 5 minutes I think you should reduce this to like 30 seconds so it doesn't step on the toes of the next flow run.

Charlie-Henry (approved):
Tested and it works 🟢 

Just a reminder to add this to our slack notifications once you get it deployed ✅

flows/atd-traffic-incident-reports/traffic-incident-reports.py
chiaberry: Good call -- I also dropped the retry to just once

mddilley (approved):
Ran into issues on my machine while running the flow (Docker error unrelated to your code), but I see that it worked for you and Charlie so I trust that the deployment will work. The code looks great so I am dropping a ✅!  🚀 🚢

frankhereford (approved):

PR #71, opened 2023-04-20
Support multiple schema of crash and related table imports in CRIS import ETL

frankhereford:
## Issue

_none_: there is not a specific issue to track this work. it was taken on while in pursuit of https://github.com/cityofaustin/atd-data-tech/issues/12053, and for a while, i thought it was the solution. 12053 is fixed in the child PR.

## Purpose

The VZ ETL, since April 1st, 2023, has been receiving two sets of CRIS crash exports, each with different crashes in them. The ETL did not expect this dual-schema delivery, and has been mixing and matching tables between them all month. ☠️ 

This patch makes the ETL CRIS-schema aware, or at least CRIS-schema capable. 

## Related PR

This is the parent/sibling of https://github.com/cityofaustin/atd-vz-data/pull/1218.

## Testing

Testing this requires setting a local development environment that offers a copy of the VZ database on port `5432` as provided by the VZ local stack as well as the prefect dockerized-agent environment which has been modified to be its own bastion host. @roseeichelmann, this will cause you particular consternation as it relies on the magic hostname `host.docker.internal` ([more info](https://austininnovation.slack.com/archives/CHZE6BC6L/p1681147029935139)). 

I think that a review of the code, in particular, the flow definition where I have adapted many of the tasks to be workable in prefect's map/reduce implementation would be a reasonable ask in terms of review. I'd welcome and enjoy a deep dive into the code with anyone as well. I'm not trying to slide in something I don't think is quality here. 

I would be _very_ happy to demo this fix. It'd be easy and quick. Please catch me on slack anytime. If anyone wants more detailed instructions on how to set this up, please let me know.

image


flows/vision-zero/cris_import/lib/sql.py
```diff
@@ -439,8 +439,8 @@ def show_changed_values(
         print(column)
         linkage_sql = " and ".join(linkage_clauses)
         sql = f"""
-            select import.{table}.{column}::text as import_{column},
-                   public.{output_map[table]}.{column}::text as public_{column}
+            select {DB_IMPORT_SCHEMA}.{table}.{column}::text as import_{column},
+                public.{output_map[table]}.{column}::text as public_{column}
```
frankhereford, line 443: I really thought I could get by without editing this file. 🫰🏻

This is a bug where I had hard-coded `import` into a SQL statement when I really wanted to interpolate the passed-in value `DB_IMPORT_SCHEMA`.

flows/vision-zero/Dockerfile
```diff
@@ -6,7 +6,7 @@ RUN apt-get update && \
     ln -fs /usr/share/zoneinfo/America/Chicago /etc/localtime && \
     dpkg-reconfigure -f noninteractive tzdata
 RUN apt-get upgrade -y
-RUN apt-get install -y vim magic-wormhole aptitude python3-pip docker.io rsync p7zip-full black poppler-utils tesseract-ocr wget postgresql-client pgloader postgresql-common libpq-dev
+RUN apt-get install -y vim magic-wormhole aptitude python3-pip docker.io rsync p7zip-full black poppler-utils tesseract-ocr wget postgresql-client pgloader postgresql-common libpq-dev openssh-server
```
frankhereford, line 9: The dockerized-agent container is actually the bastion host for the local DB in the local stack here 🪄

flows/vision-zero/Dockerfile
```diff
@@ -15,6 +15,7 @@ COPY ./prefect_bootstrap/id_rsa.pub /root/.ssh/id_rsa.pub
 COPY ./prefect_bootstrap/prefect.config.toml /root/.prefect/config.toml
 RUN chmod 600 /root/.ssh/id_rsa
 RUN chmod 644 /root/.ssh/id_rsa.pub
+RUN cp /root/.ssh/id_rsa.pub /root/.ssh/authorized_keys
```
frankhereford, line 18: Let the SSH identity of the root user on the container login into localhost via `ssh`

flows/vision-zero/cris_import/cris_import.py
```diff
@@ -40,31 +35,71 @@
...
+
+    DB_HOST = os.getenv("DB_HOST")
+    DB_USER = os.getenv("DB_USER")
+    DB_PASS = os.getenv("DB_PASS")
+    DB_NAME = os.getenv("DB_NAME")
+    DB_SSL_REQUIREMENT = os.getenv("DB_SSL_REQUIREMENT")
+
+    DB_BASTION_HOST_SSH_USERNAME = os.getenv("DB_BASTION_HOST_SSH_USERNAME")
+    DB_BASTION_HOST = os.getenv("DB_BASTION_HOST")
+    DB_RDS_HOST = os.getenv("DB_RDS_HOST")
```
frankhereford, line 102: Ugly, but useful for local testing. This is a thing I won't miss about the dockerized-agent. This could be more elegant.

flows/vision-zero/cris_import/cris_import.py
```diff
@@ -113,7 +148,7 @@ def download_extract_archives():
     # complete successfully.
     # see: https://www.gnu.org/software/libc/manual/html_node/Exit-Status.html
     if rsync.returncode != 0:
-        return false
+        return False
```
frankhereford, line 151: 🤦

flows/vision-zero/cris_import/cris_import.py
```diff
@@ -311,17 +365,22 @@ def remove_trailing_carriage_returns(data_loaded_token):
         user=DB_USER, 
         password=DB_PASS, 
         dbname=DB_NAME, 
-        sslmode="require", 
+        sslmode=DB_SSL_REQUIREMENT, 
         sslrootcert="/root/rds-combined-ca-bundle.pem"
         )
 
-    columns = util.get_input_tables_and_columns(pg, DB_IMPORT_SCHEMA)
+    columns = util.get_input_tables_and_columns(pg, map_state["import_schema"])
```
frankhereford, line 372: This variable is going to get entirely written out of the app, and you'll see this change a bunch below.

flows/vision-zero/cris_import/cris_import.py
```diff
@@ -546,59 +611,186 @@ def align_records(typed_token, dry_run):
...
 
     # fmt: on
-    return True
+    return map_state
+
+
+@task(
+    name="Group CSVs into logical groups",
+)
+def group_csvs_into_logical_groups(extracted_archives, dry_run):
```
frankhereford, line 627: The root of the solution for the split-schema problem; break the CSVs into sets based on their filename prefix.

flows/vision-zero/cris_import/cris_import.py
```diff
@@ -546,59 +611,186 @@ def align_records(typed_token, dry_run):
...
+    schema = 'import_' + hashlib.md5(mapped_state["logical_group_id"].encode()).hexdigest()[:12]
+    mapped_state["import_schema"] = schema
+    print("Schema name: ", mapped_state["import_schema"])
+    return mapped_state
 
 
+@task(
+    name="Create target import schema",
+)
+def create_target_import_schema(map_state):
```
frankhereford, line 664: Because we can't have multiple imports happening out of the same `import` schema, we're going to create a schema for every logical import group.

flows/vision-zero/cris_import/cris_import.py
```diff
@@ -546,59 +611,186 @@ def align_records(typed_token, dry_run):
...
+    pg.commit()
+    cursor.close()
+    pg.close()
+
+    return map_state
+
+@task(
+    name="Clean up import schema",
+)
+def clean_up_import_schema(map_state):
```
frankhereford, line 710: Because we're creating a schema per import to stay thread safe, we need to clean them up.

flows/vision-zero/prefect_bootstrap/start_agent.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/sh
 
+/etc/init.d/ssh start
```
frankhereford, line 3: need to start an ssh server so we can self-serve as our own bastion

flows/vision-zero/prefect_agent_environment_variables.env_template
```diff
@@ -16,7 +16,6 @@ DB_HOST=
 DB_USER=
 DB_PASS=
 DB_NAME=
-DB_IMPORT_SCHEMA=
```
frankhereford, line 19: This is dynamic now, and handled in the state-chain passed down through the `.map()`'d functions

flows/vision-zero/cris_import/cris_import.py
```diff
@@ -40,31 +35,71 @@
...
+DB_USER = None
+DB_PASS = None
+DB_NAME = None
+DB_SSL_REQUIREMENT = None
+
+DB_BASTION_HOST_SSH_USERNAME = None
+DB_BASTION_HOST = None
+DB_RDS_HOST = None
+
+if True:
```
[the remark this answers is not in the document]
frankhereford: Yep, exactly. It feels easier to me when I can inspect the environment and configuration by typing `export` and even change it temporarily inside the dockerized agent instead of pushing up to prefect and pulling it back down at run-time. Doing this more elegantly is something I'm going to touch on in our airflow demo today, for sure.

chiaberry (approved):
Read and reviewed the code, I was tripped up trying to figure out where `map_state` was defined, but then I saw that i was missing `.map` on the task invocations (changing the github view to just view the file and get the color coding syntax in all its glory helped)

The PR says the submodule updated 126 files, was that just pointing the checkout to the master branch? Or were there other changes?

chiaberry:
Just want to be clear that I didn't run any of the code, but trust that while you were testing the updates, you could see if it was working with the split schemas or not. I'm still trying to wrap my head around 
> the prefect dockerized-agent environment which has been modified to be its own bastion host.
but that doesnt prevent the code from being used to unblock the vision zero team and fix the April data

frankhereford:
@chiaberry quoted my comment here:

> the prefect dockerized-agent environment which has been modified to be its own bastion host.

When we installed the bastion host to broker access to the production VZ DB, we enhanced the ETL script to use it as a SSH tunnel. In doing so, we ended up with a ETL script that expects to connect to the DB via an intermediary machine it can SSH into. When working in the local stack, this requirement is still there, so the ETL SSH's into its own host and then connects on to the local database. It's a hoop to jump through to simulate the production setup.

It's like `ssh -L 5432:host.docker.internal:5432 localhost`.

flows/vision-zero/cris_import/cris_import.py
```diff
@@ -546,57 +610,184 @@ def align_records(typed_token, dry_run):
...
 
     # fmt: on
-    return True
+    return map_state
+
+
+@task(
+    name="Group CSVs into logical groups",
+)
+def group_csvs_into_logical_groups(extracted_archives, dry_run):
```
mddilley, line 626: Curious what your thoughts about having a known list of schemas (2018, 2023) that we check for here, and, if there is a new one, then raise an error? I don't imagine a new one coming along soon, but who knows!? Could be a way to catch a change in the future.

It also could be good to drop a comment of the known csv filename like `extract_2023_20230419092605_crash_20230418_HAYSTRAVISWILLIAMSON.csv` somewhere as an example.

flows/vision-zero/prefect_bootstrap/start_agent.sh
mddilley: This is cool for sure. Could an alternative to this be creating specific DB connect utils for local and production use?  An on/off switch of sorts for the manner of connection (one that tunnels and one that does not)? Just a thought for the next iteration.

mddilley (approved):
I did not test this locally but I know that you have tested this as you worked through it, and it can be tweaked as needed. Thank you for updating this code to handle this hard-to-navigate rollout of changes. I read through the updates thoroughly, and it makes sense to me!

flows/vision-zero/cris_import/cris_import.py
frankhereford: Yea, I think that is a super good idea. I'm going to spin off an issue for that right now.

patrickm02L (approved):
Thanks for the demo 🚀

PR #73, opened 2023-04-20
Patch to VZ Import Process

frankhereford:
# 🤦

I forgot to push a last commit into the branch with all the updates to the VZ CRIS importer.

## Intent

This PR intends to correct the key used to access the name of the import schema into which we load data from CRIS.


mddilley (approved):
🚀🚀🚀 Sorry again for overlooking this in the review this morning! 😅 Glad to see that the backfill was a success as well! 🙏

patrickm02L (approved):

chiaberry (approved):

PR #74, opened 2023-04-20
Revert "Merge branch 'atd-data03-dockerized-agent-deployment' into main"

frankhereford:
## Intent

This reverts commit d73c7073a33eb4e98bec9a70a543ec2c73d19a4a, reversing changes made to a7ec1bb737a4e56f747ad8e16f03718784953abc.

## Cause

I accidentally merged https://github.com/cityofaustin/atd-prefect/pull/55 today while thinking I was looking at a different repo. It was unwise of me to have left that PR out there, just waiting for me to make this mistake.

## Future

When this makes it into `main`, eventually, I will need to merge `main` into `atd-data03-dockerized-agent-deployment`. This will remove the changes that make production run, but I will use this PR to make them back and commit them, once again, into `atd-data03-dockerized-agent-deployment`. That will bring us back to the pre-error state. We have a good shot at actually retiring the dockerized-agent setup before this needs to happen too.

## Testing

Not really applicable here, this is best reviewed by looking at the changes in [#55](https://github.com/cityofaustin/atd-prefect/pull/55/files) and the changes in [this PR ](https://github.com/cityofaustin/atd-prefect/pull/55/files) side by side. Having the [state of undo-merge-of-pr-55](https://github.com/cityofaustin/atd-prefect/tree/undo-merge-of-pr-55/flows/vision-zero/cris_import) in a third pane is helpful, as you can check to see if a change is already in `main` and does not need reversion by this PR.

You'll notice that this PR is not the exact inverse of #55. I believe that this is the result of me merging `main` into `atd-data03-dockerized-agent-deployment` as part of my deployment work today for the VZ ETL. However, `git` is able to sort this out, and this PR was made entirely by using the one command `git revert -m 1 d73c7073a33eb4e98bec9a70a543ec2c73d19a4a`.

I have inspected both diffs line-by-line, and with the changes that this PR includes and the state of `main` into which it will be merged, and I believe this is correct.

### Thank you all! 
Sorry about this! Lesson learned; I won't make PRs I don't want to merge unless we make it incredibly impossible to merge them.

### DIFF
--- a/flows/moped/prefect_bootstrap/start_agent.sh
+++ b/flows/moped/prefect_bootstrap/start_agent.sh
@@ -6,4 +6,4 @@ dask-scheduler --dashboard-address :8787 &
 dask-worker tcp://prefect-agent:8786 &
 dask-worker tcp://prefect-agent:8786 &
 
-prefect agent local start -l atd-data03 -l moped -l production --import-path /root/test_instance_deployment/atd-moped/moped-etl/prefect
+prefect agent local start -l moped --import-path /root/test_instance_deployment/atd-moped/moped-etl/prefect
--- a/flows/vision-zero/cris_import/cris_import.py
+++ b/flows/vision-zero/cris_import/cris_import.py
@@ -35,7 +35,6 @@ from process.helpers_import import (
     insert_crash_change_template as insert_change_template,
 )
 
-
 SFTP_ENDPOINT = None
 ZIP_PASSWORD = None
 VZ_ETL_LOCATION = None
@@ -101,7 +100,6 @@ else:
     DB_BASTION_HOST_SSH_USERNAME = os.getenv("DB_BASTION_HOST_SSH_USERNAME")
     DB_BASTION_HOST = os.getenv("DB_BASTION_HOST")
     DB_RDS_HOST = os.getenv("DB_RDS_HOST")
-
 
 # Set up slack fail handler
 handler = slack_notifier(only_states=[Failed, TriggerFailed, Retrying])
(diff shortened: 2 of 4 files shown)


mddilley (commented):
I've reviewed the code and feel like the everything is matching up. I just have one question that I know that I've asked before, but why are there different states of this code that need to be held on to? It is probably that I've lost my understanding of how this works. Are the different code versions for dockerized agent vs. non-dockerized agent or different environments or something else?

Ready to turn this into an approval but I felt like I was missing that last bit of understanding.

frankhereford:
@mddilley wrote:

> why are there different states of this code that need to be held on to?

> Are the different code versions for dockerized agent vs. non-dockerized agent or different environments or something else?

#### Mike, these are great questions. To your first one:

The branch `atd-data03-dockerized-agent-deployment` exists to exactly represent the state of the code that is running on `atd-data03` as the dockerized-agents on that machine for moped & VZ. This branch contains the particular configuration changes that need to be made to this repo's `main` branch so that it runs in production. 

I would attribute to not being in a branch name of `production` and instead having this particular name because there can be as many prefect agents running on as many machines as we want, so I went with something much more specific as a name.

For a while, I was running the dockerized-agents on atd-data03 by checking out main and then making the required changes without checking them in. We conferred as a group, and we agreed that it was less good to leave those changes out there, just waiting to be lost, and instead opted created a feature branch and store them in there. That's when this `atd-data03-dockerized-agent-deployment` was created.  This would be a branch that was never intended to be merged into main. This is why having that PR which made it easy to _see_ the changes but also too easy to _merge_ the changes.

#### To your second question:

This branch is the combination of `main` and the deployment configuration required to run the Moped and VZ dockerized-agents. The prefect repo has multiple deployments over multiple versions of prefect in it, so it's certainly not clear what runs what, but this particular one is the short-for-the-world dockerized agent deploys.

I hope this answers your questions, and I'd be super eager to follow up or answer anything that would be helpful.

chiaberry (approved):

mddilley:
@frankhereford Thank you for the thorough explanation. That totally cleared up things up. 🙏

mddilley (approved):

PR #75, opened 2023-04-28
Parking Data Task-Per-Command

Charlie-Henry:
Was getting some errors from our parking data ETL and whenever it ran into an issue it would restart the whole ETL. This change moves the for loop outside the task so now each docker command is run with it's own retry settings.

### DIFF
--- a/flows/atd-parking-data/parking_data_processing.py
+++ b/flows/atd-parking-data/parking_data_processing.py
@@ -136,27 +136,27 @@ def add_command_arguments(commands, s3_env, start_date, prev_month):
 
 
 @task(
-    name="docker_commands",
+    name="docker_command",
     retries=3,
     retry_delay_seconds=timedelta(minutes=2).seconds,
+    task_run_name="Docker Command: {command}",  # note this is not wrong, you don't use an f-string here.
 )
-def docker_commands(docker_env, environment_variables, commands, logger):
-    for c in commands:
-        response = (
-            docker.from_env()
-            .containers.run(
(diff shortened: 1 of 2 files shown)


flows/atd-parking-data/parking_data_processing.py
```diff
@@ -136,27 +136,27 @@ def add_command_arguments(commands, s3_env, start_date, prev_month):
 
 
 @task(
-    name="docker_commands",
+    name="docker_command",
     retries=3,
     retry_delay_seconds=timedelta(minutes=2).seconds,
+    task_run_name="Docker Command: {command}",  # note this is not wrong, you don't use an f-string here.
```
Charlie-Henry, line 142: Will name our task based command supplied 👍

chiaberry (approved):
i like it! will the docker_env go back to production after you merge it, or is it a moot point because production and latest are the same?

Charlie-Henry:
@chiaberry yup will have to do that, it's running with my PR for parking data right now 👍

frankhereford (approved):
# 🐳🚀

This looks great. I have not attempted to run the code, but I have read it, and I think this is a elegant approach to fit more into the ETL orchestration regime. 

Thanks for doing this!

mddilley (approved):
I read the code but did not test. Thanks for updating this! 🙏 I didn't think about how Prefect would loop all the commands if any of them failed in my first look at this code. This PR is perfect to help watch out for things like this in the future. 🚢 it!

